PR #1, opened 2016-04-20
Add date to tagger object

benrudolph:
@esoergel this'll fix the tag commit issue. it silently fails if you don't include a date...

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -145,6 +145,7 @@ class DeployMetadata(object):
             tagger={
                 'name': user.name,
                 'email': user.email,
+                'date': datetime.datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ'),
             }
         )
         self._deploy_tag = tag_name


fab/fabfile.py
```diff
@@ -145,6 +145,7 @@ def tag_commit(self):
             tagger={
                 'name': user.name,
                 'email': user.email,
+                'date': datetime.datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ'),
```
esoergel, line 148: How'd you figure out what format to use?  [This](https://github.com/sigmavirus24/github3.py/blob/500138a601f65ef3f26b32c17ab7861963933207/tests/integration/test_repos_repo.py#L422-L426) is the closest thing to documentation I was able to find.  `datetime` objects also have an `isoformat()` method that might work here.

esoergel:
That sounds fun.  Thanks for the quick fix!

fab/fabfile.py
benrudolph, line 148: this cued me off: https://developer.github.com/v3/git/tags/#create-a-tag-object. it was super finicky with the format. i'll try out `isoformat` -- it has to exactly match what they specify

esoergel, line 148: Nice, makes sense that it'd be github that cares.  No need to update if this is working, I meant that more as an FYI.

PR #3, opened 2016-04-22
default email if none exists

benrudolph:
@millerdev this just fakes the email if none exists. also it uses login instead of name

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -143,8 +143,8 @@ class DeployMetadata(object):
             sha=self.deploy_ref,
             obj_type='commit',
             tagger={
-                'name': user.name,
-                'email': user.email,
+                'name': user.login,
+                'email': user.email or '{}@dimagi.com'.format(user.login),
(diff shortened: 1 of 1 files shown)


fab/fabfile.py
```diff
@@ -143,8 +143,8 @@ def tag_commit(self):
             sha=self.deploy_ref,
             obj_type='commit',
             tagger={
-                'name': user.name,
-                'email': user.email,
+                'name': user.login,
```
millerdev, line 146: was `user.name` null too? If not it might be better to use that or possibly `user.name or user.login` here.

benrudolph, line 146: i noticed some people don't have a name: https://github.com/gcapalbo. and this hardly matters cause you can't even see who made the tag in the UI: https://github.com/dimagi/commcare-hq/releases. i think the important thing is that it doesn't error and we can see the diff between tags

millerdev, line 146: :+1:

PR #4, opened 2016-04-22
Do not double define @roles

benrudolph:
@millerdev @biyeun i think this should fix the looping restart services. the `@roles` decorator was defined twice on restart services which i think caused it to execute the restart twice. tried it on staging and it didn't loop. also explains why it worked fine during deploy

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -1018,11 +1018,7 @@ def restart_services():
                            '{env.environment}?'.format(env=env), default=False):
         utils.abort('Task aborted.')
 
-    @roles(env.supervisor_roles)
-    def _inner():
-        silent_services_restart(use_current_release=True)
-
-    execute(_inner)
+    silent_services_restart(use_current_release=True)
 
 
 def silent_services_restart(use_current_release=False):


millerdev:
yay for removing unnecessary lines of code!

although this theory doesn't quite explain why it [looped 5 times for me yesterday](https://dimagi.slack.com/files/daniel/F12MP065T/fab_staging_restart_services.txt). maybe I'm not understanding something?

snopoke:
+1 also had multiple (more than 2) loops for deploy

On 22 April 2016 at 14:35, Daniel Miller notifications@github.com wrote:

> yay for removing unnecessary lines of code!
> 
> although this theory doesn't quite explain why it looped 5 times for me
> yesterday
> https://dimagi.slack.com/files/daniel/F12MP065T/fab_staging_restart_services.txt.
> maybe I'm not understanding something?
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> https://github.com/dimagi/commcare-hq-deploy/pull/4#issuecomment-213407474

## 

Simon Kelly
Senior Engineer | Dimagi South Africa

benrudolph:
huh that's strange. this is all i could see that might be causing the issue. i can't reproduce it when i try to restart services

PR #11, opened 2016-04-25
serial webworker restart

benrudolph:
@snopoke this makes it so webworkers are explicitly restarted serially. this will avoid downtime from shutting down all our webworkers at once, but add to deploy time. with captain pretty much up running, i think this is a worthwhile tradeoff @millerdev i know you also had some opinions on this

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -1029,7 +1029,7 @@ def silent_services_restart(use_current_release=False):
     Restarts services and sets the in progress flag so that pingdom doesn't yell falsely
     """
     execute(set_in_progress_flag, use_current_release)
-    execute(_restart_services)
+    execute(_restart_all_except_webworkers)
     execute(_restart_webworkers)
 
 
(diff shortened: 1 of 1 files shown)


fab/fabfile.py
```diff
@@ -1045,6 +1044,18 @@ def services_restart():
     _supervisor_command('start all')
 
 
+@roles(ROLES_DJANGO)
+@serial
```
benrudolph, line 1048: this is the default, but it's very easy to do a pass on this file and try to parallelize everything, so i'm explicitly marking as serial

fab/fabfile.py
```diff
@@ -1045,6 +1044,18 @@ def services_restart():
...
+def _restart_webworkers():
+    _services_restart()
+
+
+@roles(set(ROLES_ALL_SERVICES) - set(ROLES_DJANGO))
+@parallel
+def _restart_services():
+    _services_restart()
```
millerdev, line 1056: It's confusing to have both `_restart_services` and `_services_restart` without documentation to explain the difference (preferably the difference would be clear in the name). Suggested rename:
`_services_restart` -> `_restart_all_except_webworkers`

benrudolph, line 1056: ah good point

benrudolph, line 1056: https://github.com/dimagi/commcare-hq-deploy/pull/11/commits/b795b47a5387c5bddf878eb502887d689f22465e

PR #14, opened 2016-05-03
Add old riak machines to inventory to make cluster

esoergel:
@nickpell

### DIFF
--- a/fab/inventory/staging-va
+++ b/fab/inventory/staging-va
@@ -43,5 +43,7 @@ hqdb0-staging.internal-va.commcarehq.org
 hqdb0-staging.internal-va.commcarehq.org
 
 [riakcs]
+hqdjango4-staging.internal.commcarehq.org
+hqdjango5-staging.internal.commcarehq.org
 hqdjango0-staging.internal-va.commcarehq.org
 hqdjango1-staging.internal-va.commcarehq.org


fab/inventory/staging-va
```diff
@@ -43,5 +43,7 @@ hqdb0-staging.internal-va.commcarehq.org
 hqdb0-staging.internal-va.commcarehq.org
 
 [riakcs]
+hqdjango4-staging.internal.commcarehq.org
```
millerdev, line 46: This will change the "control node" (which is the first node in this list) for the cluster, and will distribute authentication credentials from that machine instead. If this has never been run against the new cluster then it's fine. If it has, you might want to keep `hqdjango0-staging.internal-va.commcarehq.org` at the top.

millerdev, line 46: Actually, I'm not sure it will matter either way. The credentials will be updated on the next localsettings deployment, but the old ones will continue to work, so it shouldn't break anything.

esoergel, line 46: I think I should have put that up there before deploying riak to these machines the first time.  I had to manually bring down the riak cluster on the new machines and redeploy with these settings, so they would use `hqdjango4` as the control node and join the old cluster.

millerdev, line 46: that makes sense

PR #19, opened 2016-05-25
limit memory on swiss

emord:
@benrudolph is this what you were talking about?

cc @esoergel


benrudolph:
yup thanks! @wpride is it ok that Xss is as big as Xmx

czue:
@benrudolph `Xss` is in kilobytes so I'm pretty sure it's fine.

fyi - http://stackoverflow.com/a/4967914/8207

benrudolph:
Oops misread that :+1: thanks

PR #29, opened 2016-06-07
Move old staging machines to use tsung

emord:
Doesn't include any touchforms machine (shouldn't be needed for benchmarking) and will install couch to hqdb0

buddy @proteusvacuum

### DIFF
--- a/fab/inventory/tsung
+++ b/fab/inventory/tsung
@@ -1,0 +1,36 @@
+[proxy]
+hqproxy4.internal.commcarehq.org
+
+[webworkers]
+hqdjango4-staging.internal.commcarehq.org
+hqdjango5-staging.internal.commcarehq.org
+
+[couchdb]
+hqdb0-staging.internal.commcarehq.org
+
+[postgresql]
+hqdb0-staging.internal.commcarehq.org
+
+[rabbitmq]
+hqdb0-staging.internal.commcarehq.org
+
+[zookeeper]
+hqkafka0-staging.internal.commcarehq.org
+
+[kafka]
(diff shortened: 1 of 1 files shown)


fab/inventory/tsung
```diff
@@ -0,0 +1,36 @@
...
+hqdb0-staging.internal.commcarehq.org
+
+[pillowtop]
+hqdb0-staging.internal.commcarehq.org
+
+[redis]
+hqdb0-staging.internal.commcarehq.org
+
+[elasticsearch]
+hqes1-staging.internal.commcarehq.org elasticsearch_node_name=hqes1-tsung
```
czue, line 33: do we need es for this? might save a few dollars if we put this on the db machine or one of the workers

fab/inventory/tsung
```diff
@@ -0,0 +1,36 @@
...
+hqdb0-staging.internal.commcarehq.org
+
+[rabbitmq]
+hqdb0-staging.internal.commcarehq.org
+
+[zookeeper]
+hqkafka0-staging.internal.commcarehq.org
+
+[kafka]
+hqkafka0-staging.internal.commcarehq.org kafka_broker_id=843
```
czue, line 21: similar to below, i think we do need to publish to kafka on submission but we could potentially stick it on another machine since it will be doing very little work.

czue:
no blockers

fab/inventory/tsung
snopoke, line 33: I don't think we need ES at all since we're just going to be benchmarking form submissions

PR #30, opened 2016-06-07
remove memcached

emord:
I don't think this is needed https://github.com/dimagi/commcarehq-ansible/pull/167

Also is there any reason we have preview, guinea or motech in the inventory? I don't think any of those work anymore

buddy @proteusvacuum

### DIFF
--- a/fab/inventory/staging
+++ b/fab/inventory/staging
@@ -29,9 +29,6 @@ hqtouch0-staging.internal-va.commcarehq.org
 [redis]
 hqdb1-staging.internal-va.commcarehq.org
 
-[memcached]
-hqdb1-staging.internal-va.commcarehq.org
-
 [elasticsearch]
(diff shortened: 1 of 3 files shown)


fab/inventory/staging
```diff
@@ -29,9 +29,6 @@ hqtouch0-staging.internal-va.commcarehq.org
 [redis]
 hqdb1-staging.internal-va.commcarehq.org
 
-[memcached]
```
emord, line 32: @esoergel was this needed during your migration, or was it just from copy/paste?

dannyroberts:
Haha, nice

fab/inventory/staging
esoergel, line 32: nope, just copy-pasted from the old one

czue:
> Also is there any reason we have preview, guinea or motech in the inventory?

not sure about motech, but guinea and preview can definitely be killed

PR #33, opened 2016-06-08
move kafka to db0

emord:
https://github.com/dimagi/commcare-hq-deploy/pull/29/files

@snopoke @czue 

removed ES and moved kafka over to hqdb0


czue:
:+1: 

i feel like there's a small chance not having ES might break something but we'll find out.

PR #34, opened 2016-06-09
Add new ES machines to cluster

esoergel:
This didn't go as well as planned...
@millerdev @nickpell


czue:
(put a "don't merge" label on since we've removed these from the prod cluster)

PR #35, opened 2016-06-09
Skip clean_releases until after fixing elastic

esoergel:


esoergel:
@dannyroberts @gcapalbo @czue @snopoke

czue:
good call

PR #37, opened 2016-06-09
Revert "Skip clean_releases until after fixing elastic"

esoergel:
Reverts dimagi/commcare-hq-deploy#35
That is a temporary change while we fix elasticsearch.  I'm creating this revert PR as a reminder for once that's all clear.


esoergel:
Made redundant by https://github.com/dimagi/commcare-hq-deploy/pull/46/files

PR #39, opened 2016-06-10
Revert "only run core pillows on prod until reindex is done"

czue:
Reverts dimagi/commcare-hq-deploy#38

we should merge this after reindex / memory issiues are sorted out


czue:
:+1:

PR #42, opened 2016-06-13
Revert "Revert "Use filename of application.properties""

gcapalbo:
goes with https://github.com/dimagi/commcarehq-ansible/pull/432

adding hold off on merge label so that @benrudolph can sign off
cc @wpride

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -752,9 +752,9 @@ def copy_tf_localsettings():
 def copy_formplayer_properties():
     with settings(warn_only=True):
         sudo(
-            'cp {}/submodules/formplayer/config/{}.properties '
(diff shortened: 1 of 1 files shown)


benrudolph:
i did this originally so that one day we could have an `application.properties` and a `staging.properties` that would override any non-common settings, but now realizing that's probably premature and it'd be easier to just call it `local.properties` if we ever needed it. signed off. thanks for fixing (feel free to merge, holding off due to label)

gcapalbo:
ok, removed label. can you think of anywhere else we need to update this besides https://github.com/dimagi/commcarehq-ansible/pull/432?

benrudolph:
nope, the ansible is used for the initial deployment and the fab file is used to copy it from the previous release to the current

PR #44, opened 2016-06-14
use application.properties

benrudolph:
@wpride @gcapalbo oops forgot about changing the actual supervisor task


fab/services/templates/supervisor_formplayer_spring.conf
```diff
@@ -1,6 +1,5 @@
 [program:{{ project }}-{{ environment }}-formsplayer-spring]
-command=java -jar -Xmx{{ formplayer_memory }} -Xss1024k {{ code_current}}/submodules/formplayer/build/libs/formplayer.jar --spring.config.name={{ environment }}
-user={{ sudo_user }}
```
gcapalbo, line 3: i see you also got rid of this line specifying the user, is that not needed anymore?

benrudolph, line 3: oops good catch thanks

PR #43, opened 2016-06-14
Set elasticsearch node zones

esoergel:
The first time this is used, I'll comment out the internal-va machines
@dannyroberts @snopoke @proteusvacuum 
Don't merge until this has been deployed (I'll remove the flag).

PR #47, opened 2016-06-16
Revert "Set elasticsearch node zones"

esoergel:
Reverts dimagi/commcare-hq-deploy#43
@dannyroberts 
This will make it difficult to deploy, as it'll try to connect to virginia.  Given that, I think I'm gonna use a branch called `prod-migration` which includes the new machines as I add them, and then not merge it back in until we flip the web workers.


esoergel:
self merging

PR #48, opened 2016-06-20
Small cleanup

benrudolph:
@dannyroberts i think fine to remove this yeah?

cc: @biyeun


dannyroberts:
Yeah, I think the fact that each release is in its own directory takes care of this problem (the problem of submodules we no longer use lingering after they're deleted from the repo).

PR #49, opened 2016-06-20
Medium refactor

benrudolph:
@dannyroberts doing a couple passes over the next week or so to break apart our 1400 line fabfile. was playing at making resumable deploys over the weekend and realized the code has just got to big to reason about easily

cc: @biyeun

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -43,32 +43,27 @@ from fabric.colors import blue, red, yellow, magenta
 from fabric.context_managers import settings, cd, shell_env
 from fabric.contrib import files, console
 from fabric.operations import require
+from const import (
+    ROLES_ALL_SRC,
+    ROLES_ALL_SERVICES,
+    ROLES_CELERY,
+    ROLES_PILLOWTOP,
+    ROLES_DJANGO,
+    ROLES_TOUCHFORMS,
+    ROLES_STATIC,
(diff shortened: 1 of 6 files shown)


fab/operations/db.py
```diff
@@ -0,0 +1,54 @@
+import datetime
```
benrudolph, line 1: going to start splitting out major operations like `db`, `services`, `staticfiles` etc into the `operations` module. everything in an operation module should not depend on the global state `env` and it should not define any `@task`s. at least that's the goal thus far

benrudolph:
put tag on to allow for some time to test on staging

fab/fabfile.py
```diff
@@ -1034,32 +1032,6 @@ def flip_es_aliases():
         sudo('%(virtualenv_root)s/bin/python manage.py ptop_es_manage --flip_all_aliases' % env)
 
 
-@roles(ROLES_DJANGO)
-@parallel
-def update_manifest(save=False, soft=False, use_current_release=False):
```
benrudolph, line 1117: this hasn't been necessary since we switched to atomic deploys

dannyroberts:
I am very pro this concept. Let me know when you've had a chance to test a bit / you want me to take a closer look and merge

benrudolph:
closing in favor of: https://github.com/dimagi/commcare-hq-deploy/pull/53

PR #50, opened 2016-06-21
Remove unnecessary update_manifest

benrudolph:
@biyeun @dannyroberts this is no longer needed now that we're on atomic deploys


dannyroberts:
@benrudolph can you explain why? I don't see the connection I believe you. (As in, why did it use to be necessary, and why is it now not necessary?)

benrudolph:
nvm, talked this over with biyeun and convinced myself it wasn't necessary. went to write it out, and then convinced myself it was necessary! thanks for making me think again about it. happy to talk about it though

PR #53, opened 2016-06-23
Refactor to operations

benrudolph:
@dannyroberts this runs without error on staging. will ping back when ready for review

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -58,7 +58,12 @@ from const import (
     RELEASE_RECORD,
     RSYNC_EXCLUDE,
 )
-from operations import db, staticfiles, supervisor
+from operations import (
+    db,
+    staticfiles,
+    supervisor,
+    formplayer,
+)
 
 
 if env.ssh_config_path and os.path.isfile(os.path.expanduser(env.ssh_config_path)):
@@ -517,7 +522,7 @@ def _deploy_without_asking():
 
         supervisor.set_supervisor_config()
 
-        _execute_with_timing(build_formplayer)
+        _execute_with_timing(formplayer.build_formplayer)
 
         if all(execute(_migrations_exist).values()):
             _execute_with_timing(_stop_pillows)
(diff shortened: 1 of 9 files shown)


dannyroberts:
(Assuming that "will ping back when ready for review" means it's not yet ready for review. Will look when you ping me.)

benrudolph:
Ok this is currently on staging. Other commands tested:
- [x] `fab staging restart_services`
- [x] `fab staging preindex_views`
- [x] `fab staging set_supervisor_conf`
- [x] `fab staging force_update_static`
- [x] `fab staging webworkers supervisorctl:"restart all"`

I'd say this is ready to review and happy to go through it with you. Don't think it's bad to leave on staging for a bit either

benrudolph:
@dannyroberts still easiest commit by commit i think

fab/operations/db.py
```diff
@@ -0,0 +1,96 @@
...
+            'echo "{virtualenv_root}/bin/python '
+            '{code_root}/manage.py preindex_everything '
+            '8 {user}" --mail | at -t `date -d "5 seconds" '
+            '+%m%d%H%M.%S`'
+        ).format(
+            virtualenv_root=env.virtualenv_root,
+            code_root=env.code_root,
+            user=env.user,
+        ))
+        # TODO: ask abou this staticfiles.version_static()
```
benrudolph: i don't think this is necessary and i think it got here by mistake

dannyroberts:
Hey @benrudolph, this looks good to me! Happy to merge whenever you're ready

fab/fabfile.py
```diff
@@ -510,6 +510,7 @@ def _deploy_without_asking():
         execute_with_timing(db.ensure_preindex_completion)
 
         # handle static files
+        execute_with_timing(staticfiles.prime_version_static)
```
dannyroberts, line 513: You really want to put this before `ensure_preindex_completion` if you want to keep the pipelining aspect of it.

dannyroberts, line 513: As in, `preindex_views` exists immediately, and `ensure_preindex_completion` is basically a `wait()` that will take longer to join than `prime_version_static` will to run

dannyroberts, line 513: I'd advocate for this change but I'll merge this now and let you make it later

benrudolph:
@dimagi/team-commcare-hq hey guys, this was a fairly large change to the deploy so feel free to ping me if you see a weird crash. deploys on staging went through fine for me, but it's hard to validate all the commands since we have no testing infrastructure yet for fab.

if you are thinking of adding/modifying the deploy code, i'm working on enforcing this paradigm:

`fab/operations` - these files will include "operations" which are sets of commands to be run on certain machines. you should see `@roles` in these files, but no `@task` in these files

`fab/fabfile.py` - these should _just_ be tasks. we are not at that state yet, but it'll make our deploy code easier to reason about. tasks will be made up of a composition of "operations". for sanity, tasks should never call other tasks

czue:
thanks @benrudolph this is great. what do you think about basically copying that comment into a readme so it's clear in the repo what the split is?

benrudolph:
👍 made a note to do that

PR #56, opened 2016-06-27
promote VA es nodes in inventory

snopoke:
@dannyroberts @esoergel 
so that `groups.elasticsearch.0` points to a VA node


dannyroberts:
Oops, made my own version of this just now: https://github.com/dimagi/commcare-hq-deploy/pull/57

PR #59, opened 2016-06-28
move prime_version_static

benrudolph:
@dannyroberts


dannyroberts:
@benrudolph and I talked and it turns out I was wrong, and that resource_static is super fast (10 seconds). Still, this change saves 10 seconds, so why not.

PR #61, opened 2016-06-29
refactor operations 2

benrudolph:
@dannyroberts one more of these guys, and i'll be happy. best read 🐟 . resumable deploys is on my list for PIDs and i really really needed to organize this a bit. 1400 line fabfile down to 645 lines. ran a staging deploy successfully. gonna test out a few more things, but should be ready for review.

cc: @proteusvacuum

PR #62, opened 2016-07-01
Resume deploys

benrudolph:
@dannyroberts this gives you the ability to resume deploys if they've crashed. I went with using pickle instead of redis since it's easier to store complex objects with it.
screen shot 2016-07-01 at 1 35 11 pm
best read 🐟 but could probably be either since it's fairly small. note this is a PR that goes into https://github.com/dimagi/commcare-hq-deploy/pull/61

cc: @proteusvacuum @sravfeyn

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -560,10 +560,16 @@ def awesome_deploy(confirm="yes", resume='no'):
         utils.abort('Deployment aborted.')
 
     if resume == 'yes':
-        env.update(retrieve_cached_deploy_env())
+        try:
+            cached_payload = retrieve_cached_deploy_env()
+            checkpoint_index = retrieve_cached_deploy_checkpoint()
+        except Exception:
+            print red('Unable to resume deploy, please start anew')
+            raise
+        env.update(cached_payload)
         env.resume = True
-        env.checkpoint_index = retrieve_cached_deploy_checkpoint() or 0
-        print magenta('You are about to resume the deploy in {}'.format(env.code_current))
+        env.checkpoint_index = checkpoint_index or 0
+        print magenta('You are about to resume the deploy in {}'.format(env.code_root))
 
     if datetime.datetime.now().isoweekday() == 5:
         warning_message = 'Friday'
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -33,7 +33,7 @@ from distutils.util import strtobool
 
 from fabric import utils
 from fabric.api import run, roles, execute, task, sudo, env, parallel
-from fabric.colors import blue, red
+from fabric.colors import blue, red, magenta
 from fabric.context_managers import cd
 from fabric.contrib import files, console
 from fabric.operations import require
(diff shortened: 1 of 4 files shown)


snopoke:
this looks super cool!

fab/fabfile.py
```diff
@@ -372,37 +372,30 @@ def conditionally_stop_pillows_and_celery_during_migrate():
...
+        release.clean_releases,
+    ]
 
-        # hard update of manifest.json since we're about to force restart
-        # all services
-        execute_with_timing(staticfiles.update_manifest)
-        execute_with_timing(release.clean_releases)
+    try:
+        for command in commands:
+            execute_with_timing(command)
```
esoergel: :heart_eyes:

benrudolph:
😀

fab/utils.py
```diff
@@ -102,11 +121,38 @@ def _get_github():
...
+def retrieve_cached_deploy_checkpoint():
+    filename = os.path.join(PROJECT_ROOT, CACHED_DEPLOY_CHECKPOINT_FILENAME)
+    return _retrieve_cached(filename)
+
+
+def _retrieve_cached(filename):
+    with open(filename, 'r') as f:
+        try:
+            return pickle.load(f)
+        except Exception:
```
esoergel: Do you actually want to catch this exception?  I can't imagine we'd expect the pickled file to be corrupt or otherwise unreadable, might be good to fail hard if this occurs.

fab/utils.py
```diff
@@ -102,11 +121,38 @@ def _get_github():
...
+
+
+def retrieve_cached_deploy_env():
+    filename = os.path.join(PROJECT_ROOT, CACHED_DEPLOY_ENV_FILENAME)
+    return _retrieve_cached(filename)
+
+
+def retrieve_cached_deploy_checkpoint():
+    filename = os.path.join(PROJECT_ROOT, CACHED_DEPLOY_CHECKPOINT_FILENAME)
+    return _retrieve_cached(filename)
```
esoergel, line 155: What if the file doesn't exist?

fab/utils.py
benrudolph: essentially if we can't read it or it's not there for whatever reason it'll just start the deploy from scratch

benrudolph: (offlined, going to fail hard)

fab/utils.py
benrudolph, line 155: https://github.com/dimagi/commcare-hq-deploy/pull/62/commits/7900f6e0e3fcbd23591c2046fd6a0fd982c4fb01

screen shot 2016-07-04 at 1 14 58 pm

fab/fabfile.py
```diff
@@ -560,10 +560,16 @@ def awesome_deploy(confirm="yes", resume='no'):
         utils.abort('Deployment aborted.')
 
     if resume == 'yes':
-        env.update(retrieve_cached_deploy_env())
+        try:
+            cached_payload = retrieve_cached_deploy_env()
+            checkpoint_index = retrieve_cached_deploy_checkpoint()
+        except Exception:
+            print red('Unable to resume deploy, please start anew')
```
esoergel, line 567: might want to add in a suggestion to delete the `.pickle` files so they can resume future deploys (or will that just be overwritten anyways)?

esoergel, line 567: maybe "start again without the resume flag"

benrudolph, line 567: 👍 (it'll be overwritten)

PR #63, opened 2016-07-01
Use abort exception

benrudolph:
@dannyroberts this'll better handle exceptions during deploy. now if a deploy fails during migrate (like a couple days ago) and celery doesn't come back up, the deploy exception handling code which restarts services will handle it correctly.

see here for more documention: http://docs.fabfile.org/en/1.11/usage/env.html#abort-exception

cc: @proteusvacuum @sravfeyn


snopoke:
nice

PR #67, opened 2016-07-08
Check inventory file for newrelic machines

esoergel:
And no longer hardcode in `environments.yml`
@benrudolph @snopoke

### DIFF
--- a/fab/operations/supervisor.py
+++ b/fab/operations/supervisor.py
@@ -4,6 +4,7 @@ import yaml
 import time
 import posixpath
 
+from ansible.inventory import InventoryParser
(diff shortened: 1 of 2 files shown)


fab/operations/supervisor.py
```diff
@@ -198,7 +199,11 @@ def _format_env(current_env, extra=None):
     ]
 
     host = current_env.get('host_string')
```
esoergel, line 201: This confused me, but I confirmed that this function _is_ called with `current_env['host_string']` equal to each of hosts.

benrudolph:
👍

snopoke:
👍 thanks

PR #71, opened 2016-07-13
remove zambia from fab

sravfeyn:
http://manage.dimagi.com/default.asp?231927#1182585
@biyeun


sravfeyn:
@dimagi/team-commcare-hq FYI, we don't actively support zambia server anymore.

PR #58, opened 2016-06-28
Add async restore queue

proteusvacuum:
https://github.com/dimagi/commcare-hq/pull/12251
@snopoke @sravfeyn

PR #74, opened 2016-07-18
Fix async restore supervisor program name

gcapalbo:
introduced with https://github.com/dimagi/commcare-hq-deploy/pull/58

(wasn't causing an issue yet since the case update worker runs on a different machine)

@dannyroberts code buddy
@proteusvacuum fyi


proteusvacuum:
oops, thanks @gcapalbo

PR #75, opened 2016-07-19
add exception message to deploy failures

sravfeyn:
Here's a proposal. Presently staging deploy doesn't get as much love and it mostly fails. The failure info and the debug info is usually only with the dev whoever successfully deploys the staging (I am assuming we don't log the deploy exceptions anywhere else). If we treat these failures like 500 emails, we can try to incrementally fix these since we don't resource fixing staging. We will get more visibility into deploy failures and their respective fixes so that someone like me can have these as resources while debugging deploying.

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -332,7 +332,7 @@ def hotfix_deploy():
     run('echo ping!')  # workaround for delayed console response
     try:
         execute(release.update_code, env.deploy_metadata.deploy_ref, True)
-    except Exception, e:
+    except Exception as e:
         execute(mail_admins, "Deploy failed", u"Exception message:\n{exc}".format(exc=e))
         # hopefully bring the server back to life
         silent_services_restart(use_current_release=True)
@@ -435,7 +435,7 @@ def _deploy_without_asking():
              "and wait for an email saying it's done. "
              "Thank you for using AWESOME DEPLOY.")
         )
-    except Exception, e:
+    except Exception as e:
         execute_with_timing(
             mail_admins,
             "Deploy to {} failed".format(env.environment),
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -68,6 +68,7 @@ from utils import (
     cache_deploy_state,
     retrieve_cached_deploy_env,
     retrieve_cached_deploy_checkpoint,
+    traceback_string,
 )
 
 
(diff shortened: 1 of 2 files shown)


fab/fabfile.py
```diff
@@ -332,8 +332,8 @@ def hotfix_deploy():
     run('echo ping!')  # workaround for delayed console response
     try:
         execute(release.update_code, env.deploy_metadata.deploy_ref, True)
-    except Exception:
-        execute(mail_admins, "Deploy failed", "You had better check the logs.")
+    except Exception, e:
```
czue: nit picky, but I think `except Exception as e` is preferred to avoid the confusion when trying to catch errors of multiple types

fab/fabfile.py
```diff
@@ -332,8 +332,8 @@ def hotfix_deploy():
...
+        execute(mail_admins, "Deploy failed", u"Exception message:\n{exc}".format(exc=e))
```
czue: would be nice to have the traceback here too. you can do something like this to get it:

``` python
exc_type, exc, tb = sys.exc_info()
traceback_string = "".join(traceback.format_tb(tb))
```

czue:
this is a great idea - thanks for suggesting it

czue:
my comments not blocking but will give you a chance to respond before merging

fab/fabfile.py
sravfeyn: I thought this traceback would be way too long for an email, but I will add it and revisit if it's too long latter.

fab/fabfile.py
```diff
@@ -426,6 +431,7 @@ def _deploy_without_asking():
     ]
 
     try:
+        raise Exception
```
czue: i think you need to remove this

fab/fabfile.py
czue: i think as long as it's in the body it's fine

fab/fabfile.py
sravfeyn: (updated)

benrudolph:
nice sravs! i was never a fan of the `"You had better check the logs."`

fab/utils.py
```diff
@@ -159,3 +161,9 @@ def retrieve_cached_deploy_checkpoint():
 def _retrieve_cached(filename):
     with open(filename, 'r') as f:
         return pickle.load(f)
+
+
+def traceback_string():
+    exc_type, exc, tb = sys.exc_info()
+    trace = "".join(traceback.format_tb(tb))
+    return u"Exception: {exc}\n Traceback:\n{trace}".format(exc=exc, t=trace)
```
millerdev, line 169: I think this is broken because `{trace}` and `t=trace`. Also, I'd like to suggest a small change:

``` python
return u"Traceback:\n{trace}{type}: {exc}".format(
    trace=trace,
    type=exc_type.__name__,
    exc=exc,
)
```

millerdev:
This is a great idea! Thanks @sravfeyn

fab/utils.py
sravfeyn, line 169: Ugh, I am sorry. https://github.com/dimagi/commcare-hq-deploy/pull/76

czue, line 169: my bad for not catching that

esoergel:
This is awesome!

PR #78, opened 2016-07-21
Update release.py

benrudolph:
@dannyroberts, wow somehow never noticed this...

### DIFF
--- a/fab/operations/release.py
+++ b/fab/operations/release.py
@@ -216,7 +216,7 @@ def copy_formplayer_properties():
 @roles(ROLES_ALL_SRC)
 def copy_components():
     if files.exists('{}/bower_components'.format(env.code_current)):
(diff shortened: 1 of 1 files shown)


benrudolph:
this is wrong (will come back to it later)

benrudolph:
i believe the only one broken is the static cache one not the other dirs

dannyroberts:
Yeah, the problem is creating the `js` and then moving the other `js` into it. The same line if the `js` dir doesn't exist means "copy the original dir to the new location" (which is what you want) instead of copy it _into_ the existing directory. In other other examples, I think `bower_components` doesn't yet exist when you `cp` the old one, so the previous command worked as intended.

benrudolph:
exactly. since this is an uncreated nested dir, removing the `mkdir` won't
work, but removing the last `js` will. thanks for catching 😀

PR #81, opened 2016-07-21
Download formplayer jar to build repo

benrudolph:
@snopoke this isn't complete or tested, but i realized that since we are downloading the formplayer jar from jenkins it actually makes no difference what the submodule commit is for HQ. i'm thinking about ripping out the formplayer submodule and just downloading the jar to some folder that contains the application props. that seem reasonable?


benrudolph:
ok this is g2g

benrudolph:
tested it on staging

PR #87, opened 2016-07-29
Use formplayer build dir

benrudolph:
@wpride @esoergel this'll switch us over to the new formplayer directory


fab/services/templates/supervisor_formplayer_spring.conf
```diff
@@ -1,5 +1,5 @@
 [program:{{ project }}-{{ environment }}-formsplayer-spring]
-command=java -jar -Xmx{{ formplayer_memory }} -Xss1024k {{ code_current}}/submodules/formplayer/build/libs/formplayer.jar
+command=java -jar -Xmx{{ formplayer_memory }} -Xss1024k {{ code_current}}/formplayer_build/formplayer.jar
```
benrudolph, line 2: we don't need to set the location of the config because spring will auto look in the current directory that the formplayer jar is in

esoergel:
Looks g2m when ready.

PR #97, opened 2016-08-15
removing proxy1 from proxy group

calellowitz:
@benrudolph @snopoke 
depends on https://github.com/dimagi/commcarehq-ansible/pull/504


benrudolph:
👍

PR #98, opened 2016-08-16
validate checkpoints on deploy

czue:
@benrudolph @snopoke this causes deploy to fail if checkpoints are not available in the kafka feed.

also allows you to pass `--set force=true` to bypass this.

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -87,6 +87,11 @@ if not hasattr(env, 'code_branch'):
            "You can set it with '--set code_branch=<branch>'")
     env.code_branch = 'master'
 
+
+if not hasattr(env, 'force'):
+    env.force = False  # --set force=true to override blocking warnings (e.g. stale pillow checkpoints)
+
+
 env.roledefs = {
     'django_celery': [],
     'django_app': [],
@@ -422,6 +427,7 @@ def _deploy_without_asking():
         staticfiles.prime_version_static,
         db.ensure_preindex_completion,
         # handle static files
+        db.ensure_checkpoints_safe,
         staticfiles.version_static,
         staticfiles.bower_install,
         staticfiles.npm_install,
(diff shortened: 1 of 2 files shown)


fab/fabfile.py
```diff
@@ -202,6 +207,7 @@ def staging():
         print ("using default branch of autostaging. you can override this "
                "with --set code_branch=<branch>")
 
+    env.force = True  # don't worry about kafka checkpoints on staging
```
czue, line 210: staging currently has several out of date checkpoints and figured this would be easier/less confusing than requiring people to add `force=true` every time. not sure if updating it to 30 days will fix it

fab/operations/db.py
```diff
@@ -42,6 +42,15 @@ def ensure_preindex_completion():
 
 
 @roles(ROLES_DB_ONLY)
+def ensure_checkpoints_safe():
+    extras = '--print-only' if env.force else ''
+    with cd(env.code_root):
+        sudo('{env.virtualenv_root}/bin/python manage.py validate_kafka_pillow_checkpoints {extras}'.format(
+            env=env, extras=extras
+        ))
```
snopoke: This doesn't tell people they can override with `--force=true`.  Also thinking that people might not know what the error messages mean or what to do about them. Guessing that's probably OK since it's a pretty rare occurrence but some docs somewhere could be good.

czue: yeah that's a good point. i'll see if i can make this a little better

czue: updated

fab/fabfile.py
```diff
@@ -422,6 +428,7 @@ def _deploy_without_asking():
         staticfiles.prime_version_static,
         db.ensure_preindex_completion,
         # handle static files
+        db.ensure_checkpoints_safe,
```
benrudolph, line 430: nit: should put this above the comment though i wouldn't be against removing the comment completely

benrudolph:
nice looks good

PR #100, opened 2016-08-22
Update settings for reminder rule worker

gcapalbo:
@dannyroberts


dannyroberts:
I don't necessarily expect this, but will be worth keeping an eye on the celery machines in case having more processes leads to high memory use and OOMs.

gcapalbo:
@dannyroberts yeah actually there have been some memory issues but this PR reduces the number of processes to help out with that

dannyroberts:
Sorry I read that backwards. Thanks!

PR #101, opened 2016-08-22
single quotes instead of back ticks

benrudolph:
@czue was hoping that slack would format it correctly, but turns out it's just blank 😞


esoergel:
That's a shame

PR #105, opened 2016-08-25
add server_whitelist to environment celery periodic config

dannyroberts:
will be used to make it difficult to activate celery beat
prematurely or unintentionally

@benrudolph @esoergel

### DIFF
--- a/fab/operations/supervisor.py
+++ b/fab/operations/supervisor.py
@@ -19,6 +19,7 @@ from ..const import (
     ROLES_STATIC,
     ROLES_ALL_SERVICES,
 )
+from fabric import utils
 from ..utils import execute_with_timing, get_pillow_env_config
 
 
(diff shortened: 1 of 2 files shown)


benrudolph:
👍 good call

czue:
nice

fab/operations/supervisor.py
```diff
@@ -72,14 +73,41 @@ def set_celery_supervisorconf():
...
     for queue, params in queues.items():
         for config_file in conf_files[queue]:
             _rebuild_supervisor_conf_file('make_supervisor_conf', config_file, {'celery_params': params})
 
 
+def show_periodic_server_whitelist_message_and_abort(env):
+    utils.abort(
+        "\n\n"
+        "You're seeing this message as part of an effort to reduce the chance of\n"
+        "deploying celery beat, which sends SMSes, fires off reminders,\n"
```
gcapalbo, line 90: (celery beat doesn't send sms or reminders anymore, see [here](https://github.com/dimagi/commcare-hq/blob/master/docs/messaging/reminders.rst#queueing))

fab/operations/supervisor.py
```diff
@@ -72,14 +73,41 @@ def set_celery_supervisorconf():
...
+        "and triggers forwarders, among other things, unintentionally.\n\n"
+        "For example, during a migration, you initially want to deploy everything\n"
+        "except for celery beat, because you don't want both the old and new servers\n"
+        "running celery beat; that screws up queuing, etc.\n\n"
```
gcapalbo, line 94: @dannyroberts is this in response to what happened during the virginia migration? if so the problem there was a little different and i don't think would be caught by this

gcapalbo: now that i think about it, if we also implemented the same kind of thing for these services i think we could have also avoided the virginia issue:
https://github.com/dimagi/commcare-hq-deploy/blob/55e06de71100187b7c607c4ec90917496bb51edd/fab/operations/supervisor.py#L153-L171

dannyroberts: Oh? I thought this _was_ the problem there—that celery beat was running on the new (virginia) setup while the erstwhile production setups was still running. I wasn't aware or forgot that SMS reminders are now triggered by a `run_reminder_queue` management command that replaces the polling that used to be done through celery beat. That's an analogous thing, so we want basically the same error to show up there.

That may mean slightly changing where this check lives, etc. Can you think of a full list of "things"/processes/services that should not get deployed to a pre-production setup that's currently in "mirror" mode (the way virginia was before we shut off chicago and flipped the DNS)? So far I'm thinking:
- celery beat
- run_reminder_queue
- sms queue (though this may be redundant if nothing's creating these tasks)
- reminder queue (ditto)
- pillow retry queue (not sure how these get triggered)

Anything else? Another thought to reiterate is I think it's more important to disable the thing putting things _on_ a celery queue than it is to disable the things that _process_ the queues. If you only disabled processing the celery queue then when you do the flip, there will be all these (duplicate) things in the queue that have already been processed by the previous production.

dannyroberts: To summarize, I think there's a slightly bigger/different chunk of things to include in this safety measure, and it'd be great to get your knowledge in determining those things. Once I feel like I've got the right chunk, I'll go back and redo this to cover it.

gcapalbo: yeah i was suggesting that we would want to disable the polling processes that create tasks for the reminder/sms/pillow retry queue (linked to services above).

(i think we can also take this discussion off-thread)

PR #108, opened 2016-09-05
Formplayer inventories and deploy

benrudolph:
@wpride @dannyroberts separates formplayer from touchforms for deploy as well


benrudolph:
ok deployed staging and it works

PR #109, opened 2016-09-06
No downtime web worker restart

benrudolph:
@dannyroberts this is an idea for having a no downtime web worker restart. we still experience a blip in service every time we restart. i believe this is because nginx still serves requests to the server until it realizes it's down. this would decommission the node until the restart has finished (only for clusters). we would also need to allow the `cchq` user to reload the nginx config. hasn't been tested at all, just felt inspired

### DIFF
--- a/fab/operations/supervisor.py
+++ b/fab/operations/supervisor.py
@@ -284,7 +284,7 @@ def restart_all_except_webworkers():
 def _decommission_host(host):
     files.comment(
         '/etc/nginx/sites-available/{}_commcare'.format(env.environment),
-        '^[ ]*server {}'.format(host),
+        '^[ ]*server[ ]+{}'.format(host),
         use_sudo=True,
     )
     _check_and_reload_nginx()
@@ -294,7 +294,7 @@ def _decommission_host(host):
 def _recommission_host(host):
     files.uncomment(
         '/etc/nginx/sites-available/{}_commcare'.format(env.environment),
-        host,
+        'server[ ]+{}'.format(host),
         use_sudo=True,
     )
     _check_and_reload_nginx()
--- a/fab/operations/supervisor.py
+++ b/fab/operations/supervisor.py
@@ -7,6 +7,7 @@ import posixpath
 from ansible.inventory import InventoryParser
 from fabric.api import roles, parallel, env, sudo, serial
 from fabric.context_managers import cd
+from fabric.cotrib import files
 
 from ..const import (
     ROLES_CELERY,
@@ -277,10 +278,32 @@ def restart_all_except_webworkers():
     _services_restart()
 
 
+@roles(ROLES_STATIC)
+def _decommission_host(host):
+    files.comment(
+        '/etc/nginx/sites-available/{}_commcare'.format(env.environment),
+        host,
+    )
+    sudo('nginx -s reload')
+
+
+@roles(ROLES_STATIC)
+def _recommission_host(host):
+    files.uncomment(
+        '/etc/nginx/sites-available/{}_commcare'.format(env.environment),
+        host,
+    )
+    sudo('nginx -s reload')
+
+
 @roles(ROLES_DJANGO)
 @serial
 def restart_webworkers():
+    host = env.host
+    _decommission_host(host)
     _services_restart()
+    _recommission_host(host)
+
 
 @roles(ROLES_TOUCHFORMS)
 def restart_formplayer():
--- a/fab/operations/supervisor.py
+++ b/fab/operations/supervisor.py
@@ -5,9 +5,10 @@ import time
 import posixpath
 
 from ansible.inventory import InventoryParser
-from fabric.api import roles, parallel, env, sudo, serial
+from fabric.api import roles, parallel, env, sudo, serial, execute
+from fabric.colors import red
 from fabric.context_managers import cd
-from fabric.cotrib import files
+from fabric.contrib import files
 
 from ..const import (
     ROLES_CELERY,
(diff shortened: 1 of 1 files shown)


fab/operations/supervisor.py
```diff
@@ -277,10 +278,32 @@ def restart_all_except_webworkers():
...
+    sudo('nginx -s reload')
+
+
 @roles(ROLES_DJANGO)
 @serial
 def restart_webworkers():
+    host = env.host
+    _decommission_host(host)
     _services_restart()
+    _recommission_host(host)
```
benrudolph: i went back and forth on whether or not to include this in a `finally` block. on one hand you might not want to recommission a node that has failed to restart. on the other hand you might end up in a weird state if you don't uncomment it. i'm leaning towards adding a `finally` block, but open to ideas

fab/operations/supervisor.py
```diff
@@ -277,10 +278,32 @@ def restart_all_except_webworkers():
     _services_restart()
 
 
+@roles(ROLES_STATIC)
+def _decommission_host(host):
+    files.comment(
```
snopoke, line 285: can't you just remove the link in `sites-enabled`?

fab/operations/supervisor.py
```diff
@@ -277,10 +278,32 @@ def restart_all_except_webworkers():
...
+    sudo('nginx -s reload')
+
+
+@roles(ROLES_STATIC)
+def _recommission_host(host):
+    files.uncomment(
+        '/etc/nginx/sites-available/{}_commcare'.format(env.environment),
+        host,
+    )
+    sudo('nginx -s reload')
```
snopoke: Perhaps also do `nginx -t` as a safeguard?

snopoke: Perhaps also do `nginx -t` as a safeguard?

benrudolph: good call, yes

fab/operations/supervisor.py
benrudolph: we don't want to remove the entire site, just the one host in the upstream config

benrudolph:
wow this actually works really well! if we're ok with the permission ramifications of this PR: https://github.com/dimagi/commcarehq-ansible/pull/522, then i think we'll have no downtime 🎉 i restarted webworkers multiple times while refreshing the staging page and it never went to the maintenance page

fab/operations/supervisor.py
```diff
@@ -277,10 +279,39 @@ def restart_all_except_webworkers():
...
+    sudo('nginx -s reload', shell=False, user='root')
+
+
 @roles(ROLES_DJANGO)
 @serial
 def restart_webworkers():
+    host = env.host
+    execute(_decommission_host, host)
     _services_restart()
+    execute(_recommission_host, host)
```
esoergel: very clever!

dannyroberts:
Hey @benrudolph this looks great! I think the ability to stop/start/reload nginx seems like a fine one for cchq to have. If no one disagrees, I'd say this is good to go!

benrudolph:
let's do it! i've deployed the permission changes to all the environments

fab/operations/supervisor.py
```diff
@@ -277,10 +280,49 @@ def restart_all_except_webworkers():
...
+        use_sudo=True,
+    )
+    _check_and_reload_nginx()
+
+
+@roles(ROLES_STATIC)
+def _recommission_host(host):
+    files.uncomment(
+        '/etc/nginx/sites-available/{}_commcare'.format(env.environment),
+        host,
```
snopoke: should this not use a similar regex to above?

fab/operations/supervisor.py
```diff
@@ -277,10 +280,49 @@ def restart_all_except_webworkers():
...
+
+
+def _check_and_reload_nginx():
+    sudo('nginx -t', shell=False, user='root')
+    sudo('nginx -s reload', shell=False, user='root')
+
+
+@contextmanager
+def decommissioned_host(host):
+    is_monolith = len(env.roledefs['django_app']) > 1
```
snopoke: I think this should be `not_monolith`

benrudolph: haha yes it should

fab/operations/supervisor.py
benrudolph: was more concerned about commenting out a line multiple times like `# # server host`, uncommenting doesn't have those issues so didn't think it was really necessary. can add something like `'server {}'.format(host)` though

fab/operations/supervisor.py
benrudolph: https://github.com/dimagi/commcare-hq-deploy/pull/109/commits/9c380fb9bcd6fed40250adfc273abc9f3972fac5

fab/operations/supervisor.py
benrudolph: https://github.com/dimagi/commcare-hq-deploy/pull/109/commits/6c085617515b3ad0d0c9e45bccafa5087edd95d5

benrudolph:
thanks for the feedback, updated and retested

fab/operations/supervisor.py
```diff
@@ -284,7 +284,7 @@ def restart_all_except_webworkers():
 def _decommission_host(host):
     files.comment(
         '/etc/nginx/sites-available/{}_commcare'.format(env.environment),
-        '^[ ]*server {}'.format(host),
+        '^[ ]*server[ ]+{}'.format(host),
```
dannyroberts, line 287: ha `' +'` and `'[ ]+'` are the same thing, I guess the latter's a little clearer but threw me off for a sec. 👍

PR #111, opened 2016-09-08
Try autoscale on a subset of celery workers

gcapalbo:
@dannyroberts @benrudolph 

I tested this locally and it seems to work as expected. What this does is, for a worker with a concurrency of 4 for example, instead of having 1 parent worker process and 4 child worker processes running, when there are no tasks being executed it will scale down to just 1 parent worker process and 0 child worker processes, freeing up memory.

I figured we could try this out with some of the lower priority workers first, and if all goes well then we can move all of them to use this.

So when all of these workers in this PR are idle or have little work, we're saving up to 7 processes on celery0 and 10 on celery1.


benrudolph:
nice i'm onboard

czue:
hope this works!

PR #112, opened 2016-09-08
stopasgroup | killasgroup to false for celery workers

benrudolph:
@gcapalbo @dannyroberts after some celery issue dumpster diving i think i'm getting somewhere for the mysterious `Worker exited prematurely: signal 15 (SIGTERM).`. when supervidord is set to killasgroup / stopasgroup then it will send a `SIGTERM` to all processes, including child processes. celery workers shouldn't be receiving those signals and the main process should handle it. this is just one piece of the puzzle, but if we change all our celery confs to killasgroup false then we should stop dropping tasks on restart. have yet to test it out

see `stopasgroup` and `killasgroup`: http://supervisord.org/configuration.html


gcapalbo:
nice, thanks Ben. were you able to test this out stopping a worker with a long running task?

i ask because my understanding is that when you send SIGTERM once to the worker, it will do a warm shutdown where it stops accepting new tasks and waits for all currently executing tasks (child processes) to finish before exiting process completely. so if we were to merge this as is i think we might hang on deploy if there are long-running tasks being processed.

but if we could somehow initiate the warm shutdown with this, and then still start up a new worker process (letting the warm shutdown finish on its own time), i think we would have what we want.

benrudolph:
agreed, i have the same impression as you, but i haven't been able to test it out yet. here's my theory on how we might be able to get this to work.
1. Configure celery mgmt command with  `--pidfile celery_<timestamp>.pid`
2. Set `killasgroup` to false so it signals a warm shutdown of all child processes
3. When we reload supervisor and do a [start](https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/operations/supervisor.py#L341), i _think_ that new celery process will start immediately since it uses a different PID file

what do you think? one thing i'm not sure about is if supervisor will wait until all child processes are killed or not. i was hoping that the "hanging" bit would come from the fact it was trying to start a new celery process on one that hasn't yet finished. one other thing i'm a little worried about is orphaning processes (though i think we can ensure that we clean these up with some logic in the deploy)

snopoke:
Looks good. Curious to see what supervisord does when you restart it while there are still running processes under it's control. 

We'll probably also need some way to make suer we don't end up with a bunch of zombie processes.

benrudolph:
my thought was that we'll probably have a folder of `celery_<timestamp>.pid`s on every deploy. there should only be one pid file at the start of shut down. if there are any pid files that are older, send a hard SIGQUIT to it so that the child processes do a cold shutdown

PR #113, opened 2016-09-09
errorless celery restart

benrudolph:
@gcapalbo @dannyroberts this branch is on staging and works as far as i can tell. i've been testing like this:
1. Initiate long download of an export: https://staging.commcarehq.org/a/crs-remind/data/export/custom/form/download/b6c46519a2e78e126269a26c6527904e/
2. `fab staging restart_services`. observice all pid files get removed except the main one. supervisor will hang for some timeout, then continue on and restart the rest of the services
3. start a new export of the same one. Both exports on two different celery process are being run.
4. observe that when the first export ends, the worker process all dies


benrudolph:
(requires an ansible portion to make the pid directory)

benrudolph:
thinking about this more. we may not even need all that timestamp/pid logic. i think when we SIGTERM the main process it exits and tells all the workers to exit when they're done, so we're free to start a new main process again

gcapalbo:
>  i think when we SIGTERM the main process it exits and tells all the workers to exit when they're done, so we're free to start a new main process again

@benrudolph have you tested that it works that way locally? i've observed that when you SIGTERM the parent, it waits for all children to finish before exiting itself (warm shutdown). but if you send a second SIGTERM to the parent, then it immediately kills all processes.

if you confirmed this is working then i think we're probably good, but one other thing to just keep in mind as a possible solution to this problem is to use the [pool restart](http://docs.celeryproject.org/en/latest/userguide/workers.html#pool-restart-command) as Danny had pointed out. i haven't had a chance to test that but it seems like what we want though one potential downside is we would need to figure out how to handle updates to supervisor conf files. but one potential upside is less of a chance of zombie processes. once we're comfortable with this current solution though i think we can probably try it out and see how often that happens.

fab/services/templates/supervisor_celery_repeat_record_queue.conf
```diff
@@ -1,13 +1,13 @@
 [program:{{ project }}-{{ environment }}-celery_repeat_record_queue]
 environment={% for name, value in supervisor_env_vars.items %}{{ name }}="{{ value }}"{% if not forloop.last %},{% endif %}{% endfor %}
-command={{ new_relic_command }}{{ virtualenv_current }}/bin/python {{ code_current }}/manage.py celery worker --queues=repeat_record_queue --events --loglevel=INFO --hostname={{ host_string }}_repeat_record_queue -P gevent --concurrency={{ celery_params.concurrency }}
+command=/bin/bash -c "TIMESTAMP=`date +%%s` && exec {{ new_relic_command }}{{ virtualenv_current }}/bin/python {{ code_current }}/manage.py celery worker --queues=repeat_record_queue --events --loglevel=INFO --hostname={{ host_string }}_repeat_record_queue -P gevent --concurrency={{ celery_params.concurrency }}
```
gcapalbo: we don't need the pid file for this one?

benrudolph:
>  i've observed that when you SIGTERM the parent, it waits for all children to finish before exiting itself (warm shutdown)

On staging I've observed that it waits for some set timeout  and then restarts even if the child worker does not. I found it easier to test on staging where the supervisor environment was already all setup.

> we don't need the pid file for this one?

My guess is that we actually don't need any of the pidfiles. I'm going to do some testing, but my hunch is that when you send a SIGTERM it will kill the parent process eventually and let the child processes go on their own until completion, so no need to do the timestamp pid logic. Realize there's a chance for zombie processes, but I think celery is specifically built/designed for this case

benrudolph:
Maybe once I get it setup, we can pair for a minute and convince both of ourselves that that solution will work

PR #114, opened 2016-09-13
Revert "Decommission hqriak3 on prod"

millerdev:
Re-adding hqriak3 because running out of disk space in Riak CS cluster.

@dannyroberts cc @NoahCarnahan


millerdev:
already re-added this machine to the prod riak cs cluster

PR #115, opened 2016-09-13
Swiss optimize

benrudolph:
@czue @emord this implements a few of the changes to help ease the load on swiss (http://manage.dimagi.com/default.asp?231012)

i punted on coalescing celery tasks since gio is planning on implementing `autoscale` for the rest of the works soon

cc: @gcapalbo

### DIFF
--- a/fab/pillows/swiss.yml
+++ b/fab/pillows/swiss.yml
@@ -5,3 +5,5 @@ exclude_pillows:
   - ReportCaseToElasticsearchPillow
   - ReportXFormToElasticsearchPillow
   - CaseSearchToElasticsearchPillow
+  - kafka-ucr-static
+  - kafka-ucr-main
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -116,7 +116,7 @@ swiss:
   service_blacklist:
     - supervisor_websockets.conf
     - supervisor_celery_flower.conf
-    - supervisor_formsplayer_spring.conf
+    - supervisor_formplayer_spring.conf
 
   celery_processes:
     '*':
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -113,6 +113,11 @@ swiss:
   jython_memory: "1024m"
   formplayer_memory: "1024m"
 
+  service_blacklist:
+    - supervisor_websockets.conf
+    - supervisor_celery_flower.conf
+    - supervisor_formsplayer_spring.conf
+
   celery_processes:
     '*':
       main:
--- a/fab/operations/supervisor.py
+++ b/fab/operations/supervisor.py
@@ -184,8 +184,8 @@ def set_websocket_supervisorconf():
 def _rebuild_supervisor_conf_file(conf_command, filename, params=None):
     sudo('mkdir -p {}'.format(posixpath.join(env.services, 'supervisor')))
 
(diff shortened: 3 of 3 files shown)


fab/pillows/swiss.yml
```diff
@@ -5,3 +5,5 @@ exclude_pillows:
   - ReportCaseToElasticsearchPillow
   - ReportXFormToElasticsearchPillow
   - CaseSearchToElasticsearchPillow
+  - kafka-ucr-static
+  - kafka-ucr-main
```
benrudolph, line 9: didn't see them using any ucr so figured this was fine. @czue let me know if you think that's an issue

emord: How hard would it be to get them to run these services with `pillow-key` instead of `pillow-name`, or even just one `run_ptop` service for all pillows? I doubt having all of these services separate matters for  low volume server

emord:
> i punted on coalescing celery tasks since gio is planning on implementing autoscale for the rest of the works soon

I might be missing something but I think that these two things might be separate tasks. Right now we have a separate celery worker for each celery task (at least we have separate sevices files for each of them). I think that autoscale only works on a per worker basis, so even after autoscale we would still have x number of celery processes since all of the swiss jobs have a concurrency of 1.

emord:
Looks good to me. Will let @czue confirm about UCRs

fab/pillows/swiss.yml
benrudolph: yeah i've always wondered about that as well. i don't think it'd be too too hard, but this might do the trick as here are the current processes that take up a lot of mem:

screen shot 2016-09-13 at 5 29 36 pm

kafka-ucr-static will be eliminated (#3)
formplayer.jar will be eliminated (#4)

after that list is a bunch of celery which will hopefully be eased with autoscale. lastly i think we could cut down on ES mem if we're still having troubles. i think if this doesn't help, we should definitely implement the pillow-key thing

benrudolph:
> I might be missing something but I think that these two things might be separate tasks. Right now we have a separate celery worker for each celery task (at least we have separate sevices files for each of them). I think that autoscale only works on a per worker basis, so even after autoscale we would still have x number of celery processes since all of the swiss jobs have a concurrency of 1.

Right, but from what I understood from talking with gio about this is that the actual parent worker task that spawns the child workers to do the actual task has a low memory footprint. So removing the child workers when we don't need them would be where we would save significant amounts of memory

fab/pillows/swiss.yml
czue: sounds good. also sounds fine to remove the UCR queues for now as well though wonder if this will confuse us at some point down the road...

benrudolph: yeah was wondering how sneaky this would be. one thought would be to have
some sort of error when trying to create ucr's on swiss until we remove
this (maybe a soft assert?)

benrudolph:
i think this is good to go. i've made a note to followup with a soft assert if that process isn't running on swiss and you try to build a UCR. i don't think it should block this though

emord:
image

update from swiss after this was deployed. quite a bit of breathing room

benrudolph:
🌈

czue:
:smile: :computer:

PR #117, opened 2016-09-15
Update remaining workers to use autoscale

gcapalbo:
following up from https://github.com/dimagi/commcare-hq-deploy/pull/111

it's been a week and autoscale looks stable, we haven't had issues running tasks with it. so given the memory usage gains i think it's worth it to make the rest of the workers use it too.

@benrudolph


benrudolph:
nice! the swiss will thank you 🍷

snopoke:
great! will be interesting to watch the celery alerts to see if we still get as many - hopefully not.

czue:
thanks giovanni!

PR #118, opened 2016-09-19
Bash controlled celery

benrudolph:
alright @gcapalbo third time's the charm! played around with this on staging and i think it does exactly what we'd want. you made a good point about monitoring that i haven't yet tested. just for posterity here's how it works
1. supervisor will spawn off a bash process
2. the bash process then spawns a celery process
3. when supervisor restarts, it'll send a SIGTERM to the bash process
4. the bash process will catch the SIGTERM and forward it to celery initiating a warm shutdown
5. the bash process exits, which frees up supervisor to restart a new bash process
6. the celery process is able to do a complete warm shutdown

fyi: @snopoke @czue


snopoke:
How will this work with things like exports. Will the progress bar on HQ still work?

benrudolph:
yep, everything should look like normal on the hq side

benrudolph:
deployed to staging

fab/services/templates/supervisor_celery_main.conf
```diff
@@ -1,6 +1,28 @@
...
+;
+; > When Bash is waiting for an asynchronous command via the wait
+; > built-in, the reception of a signal for which a trap has
+; > been set will cause the wait built-in to return immediately
+; > with an exit status greater than 128, immediately after which
+; > the trap is executed.
+;
+; See Section 12.2.2 of http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_12_02.html
+;
+; Effectively we've no orphaned
```
gcapalbo: minor: i think we can remove the word "no"

gcapalbo:
Ben this is great, I'm glad to see there was a simple solution to this. Were you able to test on staging and observe the warm shutdown?

I think it's worth figuring out how this might interfere with uptime monitoring though before we move to prod. Specifically, I want to make sure if there's a situation where a celery worker is in warm shutdown and it fails to restart on the new process for any reason, that we don't count that as being up.

benrudolph:
👍 i'll do some investigatin

benrudolph:
Ok was able to test both the monitoring of celery and flower. In both cases I believe they do what we want.

For Celery:
1. Created a task that lasted 20 seconds
2. Pinged the worker and received a pong
3. Initiated task and then sent celery to warm shutdown
4. Observe that ping no longer returns a response
(this is the check we do in `check_celery`)

screen shot 2016-09-20 at 12 20 45 am

For Flower:
1. Created a task that lasted 20 seconds
2. Called the flower api for all workers (observe that the worker is present)
3. Initiated task and then sent celery to warm shutdown
4. Observe that the api no longer says that the worker is up
screen shot 2016-09-20 at 12 33 39 am
(this is the check we do in `check_heartbeat`)

czue:
neat stuff.

benrudolph:
(added label have to make this an actual bash script because supervisor is have trouble parsing)

gcapalbo:
thanks Ben for investigating. Based on those tests it does look like we won't miss any checks we're already performing. But I think we might also get a false alarm every time we deploy for the `check_heartbeat` check because the hostnames on the workers are going to be different pre and post deploy, so it would report the old worker as down and the new one as up.

benrudolph:
true, but that blip is usually covered by the `DEPLOY_IN_PROGRESS` flag that gets set when we restart services (kind of cheating). i'm not exactly sure how flower determines its host list, but i'll look into that and see what potential issues would arise from that. we restart flower on deploy so i'm not too concerned about the blip during deploy. i think it might be an issue when someone restarts just the single queue

benrudolph:
ok the celery worker down notification definitely persists on restart of the main queue. only solution i can think of right now is to trigger a restart of flower when the main queue restarts

gcapalbo:
i think another solution is to just change how we do the check. i haven't thought too much about it yet but maybe we can just hard code the list of hosts to monitor and check for them using a substring match. i looked at `check_celery`, and it looks like it only makes sure that at least one worker is up which doesn't seem too useful, so maybe we can replace that check with a new one.

fab/services/templates/celery_main_bash.sh
```diff
@@ -0,0 +1,21 @@
+# Catch any TERM or INT signals and propagate
```
gcapalbo, line 1: i kind of liked it better when it was just one file to add every time we add a new worker instead of two. could we make a generic shell script that gets invoked from supervisor and passes all of the args it receives to ./manage.py celery worker?

benrudolph:
yes, good point, as soon as i wrote that i looked at our checks and i think it'd be fairly easy to do. we can even ensure the most recent one is up and the other ones are down based on the timestamp (this way we can determine zombie processes). agree check_celery is pretty useless

fab/services/templates/celery_main_bash.sh
benrudolph: yeah also would like that more

benrudolph:
@gcapalbo i added one more commit that adds the timestamp suffix so the uptime check works: https://github.com/dimagi/commcare-hq-deploy/pull/118/commits/21d1fc933b7a9b5e6de92a2a0d0db4d3db547bd7

gcapalbo:
:+1: good to merge from my side, will wait until you finish testing

benrudolph:
ok did one last deploy and tested the restart functionality and flower status monitoring. i think we're good to go once the new `check_heartbeat` check gets merged 😲

image

PR #119, opened 2016-09-20
Move case update worker to celery1

gcapalbo:
@benrudolph 
celery0 seems to have higher memory usage than celery1, and i've also seen some higher spikes in celery0's memory usage (one time i saw the main worker was taking up 50% of the memory on the machine). this will hopefully help even out the memory usage between the machines.


benrudolph:
nice sounds good

PR #123, opened 2016-09-22
Revert "Revert "Bash controlled celery""

gcapalbo:
Reverts dimagi/commcare-hq-deploy#122

@benrudolph when this was deployed the main worker didn't start because it was looking for `production_celery_main_bash.sh` that didn't exist (no *.sh file was copied into the services directory)

there's no rush on this, figured we'd just revert and you can have a look whenever you have time

### DIFF
--- a/fab/operations/supervisor.py
+++ b/fab/operations/supervisor.py
@@ -61,7 +61,7 @@ def _get_celery_queues():
 def set_celery_supervisorconf():
 
     conf_files = {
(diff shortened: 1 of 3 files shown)


benrudolph:
Huh strange. Thanks for handling and sorry about that. Was not expecting
that. I've deployed multiple times to staging

benrudolph:
(offlined with gio, we're pretty sure this is due to the fact we deployed this change on a stale fabfile. that's why there's no bug fix here)

PR #125, opened 2016-10-03
Remove celery flower from blacklist

benrudolph:
@nickpell this'll have flower run on swiss again. makes sense to add back in since we depend on it for our celery worker checks

http://manage.dimagi.com/default.asp?238637


nickpell:
👍

PR #126, opened 2016-10-04
adding db0 as india standby to get newest backup script

calellowitz:
@gcapalbo 
All the backup work is in the standby task now, so we need to designate the main db machine as the standby one to get the latest chagnes

### DIFF
--- a/fab/inventory/softlayer
+++ b/fab/inventory/softlayer
@@ -93,3 +93,6 @@ datavol_device=/dev/xvdc
 
 [stanchion:children]
 riak5
+
+[pg_standby:children]
+db0


gcapalbo:
this looks ok to me but i don't have much context. is there anyone who has more context that you'd like to review or are you good with this?

calellowitz:
im good with this. it only changes the ansible inventory file, which shouldn't have any destructive consequences. and I have been manually changing it for tests anyway. this mostly just formalizes it, but it means anytime we change postgres backups we can deploy those to india and if we need to set up a new cluster this will be ready

PR #128, opened 2016-10-07
Generalize celery bash runner

benrudolph:
@gcapalbo we've deployed a few times with the new bash runner, and things seem to be going well. i thought i'd take a stab at generalizing it for other queues to use

cc: @orangejenny


benrudolph:
(this code hasn't been run yet, need to test out on staging)

fab/services/templates/celery_bash_runner.sh
```diff
@@ -15,7 +15,27 @@
 # the Celery process to do a warm shutdown and we are
 # free to start another bash process under supervisor.
 trap 'kill -TERM $PID' TERM INT
+
+HOSTNAME=""
+ARGS=""
+for i in "$@"
```
benrudolph, line 21: i hate this loop, but there didn't seem to be a great way to do this. i feel like it's probably OK for this isolated scenario

gcapalbo:
looks good to me, good to merge once you're done testing

benrudolph:
this had been working on staging for the last day and has been deployed multiple times. 
screen shot 2016-10-11 at 2 06 12 pm
the proper args are propagating and the main queue is responding to new tasks. think this is good to go

benrudolph:
cool! next up on the chopping block is the ucr queue to get rid of these guys: `Task corehq.apps.userreports.tasks.rebuild_indicators WorkerLostError('Worker exited prematurely: signal 15 (SIGTERM)`

PR #130, opened 2016-10-13
Bash runner for UCR queue

benrudolph:
@czue @gcapalbo how do you guys feel about converting the ucr queue to use the bash runner


czue:
good by me - my understanding is that we don't know of any major downside to doing this right?

gcapalbo:
:+1: i'm good with doing this for all the workers

benrudolph:
cool, one thing to consider though is that we may have a lot more running tasks. if we don't have logic that stops users from spawning off more than one ucr rebuild, then it's possible that we can queue up a large number of these tasks. though hard limit for any celery task right now is 2 days

PR #131, opened 2016-10-13
Add swap to riakcs hosts

millerdev:
Riak has been running well since I added swap during the dev summit. No OOM crashes and no noticeable performance hit since an initial period of slowness just after enabling swap. Maybe riak's memory usage patterns paired with our sparsely accessed data set actually works well with swap? If so it's a cheap way to feed a memory hungry beast.

Ansible PR coming up next.

@snopoke @czue cc @kaapstorm @proteusvacuum

### DIFF
--- a/fab/inventory/production
+++ b/fab/inventory/production
@@ -65,6 +65,7 @@ hqriak7.internal-va.commcarehq.org
 
 [riakcs:vars]
 datavol_device=/dev/xvdb
+swap_size=4G
 
 [stanchion]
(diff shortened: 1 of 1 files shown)


czue:
:+1: makes sense to me. maybe we can also get rid of the restarter if this fixes that

millerdev:
> maybe we can also get rid of the restarter if this fixes that

Interesting idea. I'll have to think about that.

PR #134, opened 2016-10-17
Formplayer JMX

benrudolph:
@snopoke @wpride this'll enable JMX monitoring for formplayer. Don't merge yet because currently the process can't read the jmx password file and causes it to crash. Once the permissions are changed though I'm able to connect to it locally through JConsole


fab/services/templates/supervisor_formplayer_spring.conf
```diff
@@ -1,5 +1,5 @@
 [program:{{ project }}-{{ environment }}-formsplayer-spring]
-command=java -jar -Xmx{{ formplayer_memory }} -Xss1024k {{ code_current}}/formplayer_build/formplayer.jar
+command=java -jar  -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote=true -Dcom.sun.management.jmxremote.port=9999 -Xmx{{ formplayer_memory }} -Xss1024k {{ code_current}}/formplayer_build/formplayer.jar
```
benrudolph, line 2: I don't think it matters that we don't authenticate/ssl since we don't open that port and it's behind vpn

benrudolph:
This is all set to go

PR #136, opened 2016-10-19
Add some logging to bash script

benrudolph:
@gcapalbo did some digging and nothing obvious or repeatable is coming up. I've added some logging here to make it easier to go back and see what happened

### DIFF
--- a/fab/services/templates/celery_bash_runner.sh
+++ b/fab/services/templates/celery_bash_runner.sh
@@ -14,7 +14,7 @@
 # Effectively we've orphaned
 # the Celery process to do a warm shutdown and we are
 # free to start another bash process under supervisor.
-trap 'echo "Killing: $PID"; kill -TERM $PID; echo "Killed: $PID";' TERM INT QUIT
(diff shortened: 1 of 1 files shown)


fab/services/templates/celery_bash_runner.sh
```diff
@@ -14,7 +14,7 @@
 # Effectively we've orphaned
 # the Celery process to do a warm shutdown and we are
 # free to start another bash process under supervisor.
-trap 'kill -TERM $PID' TERM INT
+trap 'echo "Killing: $PID"; kill -TERM $PID; echo "Killed: $PID";' TERM INT QUIT
```
benrudolph: note i added QUIT here in case for some reason supervisor got anxious and called SIGQUIT

benrudolph: actually i think supervisor would send a KILL when it need to terminate. no where in the logs does it suggest that that has happened though...

benrudolph:
i get the sense we're dealing with something that's gonna be hard to track down

PR #139, opened 2016-10-27
Add kill stale workers

benrudolph:
@gcapalbo this should mitigate our issues we've been seeing around random celery workers continuing to exist. tested it on prod and staging and it worked. it shouldn't interfere with workers that have properly engaged in a warm shutdown

cc: @emord @nickpell

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -359,6 +359,14 @@ def _confirm_translated():
 
 
 @task
+def kill_stale_celery_workers():
+    """
+    Kills celery workers that failed to properly go into warm shutdown
+    """
+    execute(release.kill_stale_celery_workers)
+
+
+@task
 def deploy_formplayer():
     execute(formplayer.build_formplayer, True)
     execute(supervisor.restart_formplayer)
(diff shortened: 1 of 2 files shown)


gcapalbo:
i think this would solve the issue for when this happens on deploy but not when it happens on just restarting the supervisor service.

ideally i think we should incorporate the new shutdown procedure as our way to shutdown an old worker before starting the new one.

gcapalbo:
maybe we could make the management command shutdown only one worker, and then have the shell script run the management command when it catches the TERM signal (instead of propagating TERM to the worker)

benrudolph:
> i think this would solve the issue for when this happens on deploy but not when it happens on just restarting the supervisor service.

that's correct, not the greatest solution

> maybe we could make the management command shutdown only one worker, and then have the shell script run the management command when it catches the TERM signal (instead of propagating TERM to the worker)

that makes sense to me, i only worry that it takes quite a bit of time to run a mgmt command and it'd increase restart time. however i think the ability to have a reliable shut down is worth the wait. will look into it

PR #142, opened 2016-11-01
Delay kill

benrudolph:
@dannyroberts @gcapalbo 

if at first you don't succeed, try again in 3 minutes

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -483,7 +483,6 @@ def _deploy_without_asking():
     else:
         execute_with_timing(release.update_current)
         silent_services_restart()
-        execute(release.delay_kill_stale_celery_workers)
         execute_with_timing(release.record_successful_release)
         execute_with_timing(release.record_successful_deploy)
         clear_cached_deploy()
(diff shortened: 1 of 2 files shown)


benrudolph:
hold on! gonna make an adjustment

benrudolph:
ok 👍 this'll run it after restart...

dannyroberts (approved):
I heartily approve

benrudolph:
@gcapalbo i know you were interested in consolidating the two shutdown methods. i think i'm going to go with this to start so i can stop having to babysit every deploy and if it works i'll make a `kill_celery_worker_by_hostname` command that takes a hostname arg and we can call that management command from supervisor. i have to do some testing to make sure it goes into warm shutdown. the docs just say "graceful shutdown" which i assume is a warm shutdown: http://docs.celeryproject.org/en/latest/userguide/workers.html#remote-shutdown. but will test

gcapalbo:
hmm, i feel like this keeps getting more complicated. i hope this is just a temporary solution and not a permanent one.  my vote would still be to find the way that shuts down the workers reliably and exclusively use that as our method.

benrudolph:
agreed, i'm unfortunately in a funny state where i don't think it makes sense to rollback, but we can't keep having these pingdom down messages and getting to the bottom of it has proven difficult. it's pretty much, try this solution and hope it works since we can't reliably reproduce. definitely not going to stop trying to get it to work properly though

also the other fix where we called this originally got reverted, so this is the only time we call `kill_stale`

PR #143, opened 2016-11-01
Use different method for shutdown

benrudolph:
@gcapalbo Going to do some testing on staging, so don't merge yet, but hopefully this'll be a more sustainable solution

cc: @NoahCarnahan

### DIFF
--- a/fab/services/templates/celery_bash_runner.sh
+++ b/fab/services/templates/celery_bash_runner.sh
@@ -14,9 +14,11 @@
 # Effectively we've orphaned
 # the Celery process to do a warm shutdown and we are
 # free to start another bash process under supervisor.
-trap 'echo "Killing: $PID"; kill -TERM $PID; echo "Killed: $PID";' TERM INT
+trap 'echo "Killing: celery@${HOSTNAME} - $PID";
+    {{ virtualenv_current }}/bin/python {{ code_current }}/manage.py shutdown_celery_worker_by_hostname celery@${HOSTNAME};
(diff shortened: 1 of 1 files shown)


benrudolph:
we may want to handle the exit code 1 and do a retry or something

czue:
haha such a battle. :popcorn:

gcapalbo:
@benrudolph yeah i was going to suggest retrying on exit code 1 as you said. otherwise, looks good to me

benrudolph:
one thing we may be worried about is the memory overhead it now takes to shut down a celery worker. running a management command takes 100MB - 160MB of ram. If we have 5 of these it means we take ~0.5GB to shut down the workers. don't think it's a blocker but something to be aware of

benrudolph:
Ok this is deployed to staging and works on restart. g2m from my perspective.  let's see it works! 🔬

gcapalbo:
do you think it's worth trying to catch exit code 1 before we release? one upside to doing that is it will make us more proactive when it doesn't work

benrudolph:
sure, were you thinking something like `if (exit_code == 1) redo the same command`? guessing we don't want to make it too complex or give it the potential to become an infinite loop

gcapalbo:
yeah, though now that i think about it maybe it's better to try this out first and see how it goes. better to not make it too complicated unless we need to.

let me know what you think - i'm good to merge and try it out if you are

benrudolph:
ok sounds good to me. staging has undergone a few restarts and two deploys with no weird things showing up, merging

PR #144, opened 2016-11-16
Remove delay kill celery workers

benrudolph:
@gcapalbo i received no soft asserts since moving to the new way of killing the celery workers, so i think we may have finally won the war on this.

i searched my inbox and there has only been 2 worker exited early due to SIGTERM for the main queue since we switched to `_timestamp`ed workers (circa Oct 5th). i'm pretty sure those were due to us manually killing the queue. this also means we should officially have a no downtime deploy (though users may experience some slowness) 🎉 level unlocked

keeping kill_stale_workers around just in case

cc: @mkangia


gcapalbo:
great!

PR #146, opened 2016-11-23
disable some queues

snopoke:
@calellowitz @sravfeyn 

Disable these until the GZB users have been deleted to prevent sending out SMS's etc.

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -64,22 +64,22 @@ icds:
         server_whitelist: 10.247.24.19
       pillow_retry_queue:
         concurrency: 1
-      reminder_case_update_queue:
-        concurrency: 4
-      reminder_queue:
-        concurrency: 8
-      reminder_rule_queue:
-        concurrency: 1
+#      reminder_case_update_queue:
+#        concurrency: 4
+#      reminder_queue:
+#        concurrency: 8
+#      reminder_rule_queue:
+#        concurrency: 1
       saved_exports_queue:
         concurrency: 3
-      sms_queue:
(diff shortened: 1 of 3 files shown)


gcapalbo:
@snopoke i don't have much context here but fyi the tasks will still keep queuing up and then will be processed all at once when these workers come back online (not sure if that's intended or not)

calellowitz:
That does not sound good. we could always purge the queues before them back online. Would that work @gcapalbo ?

snopoke:
Hey, that's not intended. Basically what's going to happen is that the user
data for all users (except test users) is going to be deleted at some later
point in time.

In the mean time we need to be able to do QA on the domain without firing
reminders etc.

What approach would you suggest?

gcapalbo:
@snopoke @calellowitz there should only be one domain right? i don't have access so i can't check but i think all the reminders only send emails for the issue tracker at this point and don't send any sms. if you want those to stop, probably the easiest way to do so would be to just deactivate those reminders (can do so from the Messaging -> Reminders page).

for repeaters, i'm not sure how to stop that without deleting the forwarding rules (or introducing a one-off feature flag) but based on the domain on the india server it doesn't look like there are any forwarding rules configured.

also, let me know when/how you're going to be doing the deleting and i can let you know what messaging models (if any) need to be deleted as well.

gcapalbo:
@snopoke @calellowitz if you agree with my comments above, can we revert this?

@sravfeyn - we won't be able to test the gateway setup if this change gets deployed

PR #147, opened 2016-11-23
only run certain command once per deploy

snopoke:
With multiple DB machines e.g. NIC these tasks all get run once per host which isn't required.

Not yet tested hence the label.

@benrudolph 
buddy @biyeun 

ref: http://docs.fabfile.org/en/latest/api/core/decorators.html#fabric.decorators.runs_once


benrudolph (approved):
Nice this is simpler than i thought it'd be

czue:
that is nice

snopoke:
:white_check_mark:

PR #149, opened 2016-11-28
Do not do a double restart for monoliths

benrudolph:
@mkangia / anyone.  realized we were restarting services twice on monoliths since we don't actually group our supervisor tasks. we just restart by machine. this should also fix the celery stale worker issue on swiss

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -68,6 +68,7 @@ from utils import (
     retrieve_cached_deploy_env,
     retrieve_cached_deploy_checkpoint,
     traceback_string,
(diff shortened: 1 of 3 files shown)


mkangia:
@benrudolph 👍 
notifying @nickpell as this takes care of the issue on swiss for running old code at times that he raised.

nickpell:
yay!

benrudolph:
yeah i'm pretty confident this will fix that issue. was able to reproduce the issue and this fixed. sorry it took me this long to track down. was not expecting this to be the issue!

PR #150, opened 2016-12-01
Announce deploy start for formplayer

benrudolph:
@emord / anyone


emord:
Will this send the notification on each deploy, or just when you're deploying formplayer by itself?

benrudolph:
just when formplayer is deployed by itself

PR #151, opened 2016-12-01
Run create_kafka_topics on deploy

esoergel:
This is pretty quick when it's a noop, so I don't think it's a big deal to add.  I just tested `$ time fab staging manage:create_kafka_topics` and it took 7 seconds.

I'm not sure exactly when is the most appropriate place to add this, and would be willing to hear second opinions.


benrudolph (approved):

esoergel:
@NoahCarnahan @czue from https://github.com/dimagi/commcare-hq/pull/13976

czue:
nice one, thanks.

PR #152, opened 2016-12-02
Apply patch

benrudolph:
@snopoke @czue this should do the trick. got it working on staging

cc: @emord

### DIFF
--- a/fab/operations/release.py
+++ b/fab/operations/release.py
@@ -286,7 +286,7 @@ def mark_keep_until(keep_days):
 @roles(ROLES_ALL_SRC)
 @parallel
 def apply_patch(filepath):
-    destination = '/home/{}/patch.patch'.format(env.user)
+    destination = '/home/{}/{}.patch'.format(env.user, env.deploy_metadata.timestamp)
     operations.put(
         filepath,
         destination,
--- a/fab/operations/release.py
+++ b/fab/operations/release.py
@@ -6,7 +6,7 @@ from fabric.api import roles, parallel, sudo, env
 from fabric.colors import red
 from fabric.context_managers import cd, settings
 from fabric.contrib import files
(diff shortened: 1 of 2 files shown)


fab/operations/release.py
```diff
@@ -281,3 +281,16 @@ def mark_keep_until(keep_days):
     until_date = (datetime.utcnow() + timedelta(days=keep_days)).strftime(DATE_FMT)
     with cd(env.code_root):
         sudo('touch {}{}'.format(KEEP_UNTIL_PREFIX, until_date))
+
+
+@roles(ROLES_ALL_SRC)
+@parallel
+def apply_patch(filepath):
+    destination = '/home/{}/patch.patch'.format(env.user)
```
czue: would it make sense to use unique names for these?

czue (approved):
nice glad this was this simple

fab/operations/release.py
benrudolph: could do, my initial logic was i didn't want a bunch of files hanging around, but now thinking about it again it'd be useful to have the date in the filename so we know when it happened

benrudolph: https://github.com/dimagi/commcare-hq-deploy/pull/152/commits/45023919d99808cd6067bc3318a81c6a8ac0cf45

czue: yeah i was thinking that if we did this multiple times we might want a record on the servers of what had happened. and there would be less likelihood of bugs related to applying the wrong patch (not sure how that would happen but was trying to be paranoid)

snopoke:
That was fast.

czue:
@benrudolph is a :1234: :racehorse:

snopoke (approved):

benrudolph:
:robot:

PR #154, opened 2016-12-10
convert-int-types-to-string patch

benrudolph:
@czue how do you feel about having it in the deploy repo? i think the patches more logically fit with environments over django apps and i'm hesitant to re-add a dependency between hq and deploy


czue (approved):
yeah i like this better, thanks

PR #157, opened 2016-12-14
use roles_control

benrudolph:
@snopoke @emord how do you feel about switching to just using a role that is one machine? the `@runs_once` would cause preindex to hang, but then as i was fixing it, i realized that it probably makes more semantic sense to just have a role with one machine then to have a role with multiple machines and have it silently fail on the others. still need to test this code

cc: @millerdev


emord:
Sounds reasonable to me. Thanks for looking into that

snopoke (approved):
:+1:

emord:
Worked great on prod. any other tests you want to do before merging?

benrudolph:
mergin'

PR #158, opened 2016-12-15
Fix mail admins

benrudolph:
@millerdev I believe the quotes were not properly escaped when sending the traceback to `mail_admins` and actually caused the failed command to be run again. this was based off looking at the traceback you sent me:

```
[hqdb1-staging.internal-va.commcarehq.org] out: usage: manage.py migrate_multi [-h] [--version] [-v {0,1,2,3}]
[hqdb1-staging.internal-va.commcarehq.org] out:                                [--settings SETTINGS] [--pythonpath PYTHONPATH]
[hqdb1-staging.internal-va.commcarehq.org] out:                                [--traceback] [--no-color] [--noinput] [--fake]
[hqdb1-staging.internal-va.commcarehq.org] out:                                [--list]
[hqdb1-staging.internal-va.commcarehq.org] out:                                [app_label] [migration_name]
[hqdb1-staging.internal-va.commcarehq.org] out: manage.py migrate_multi: error: unrecognized arguments: --slack --environment`
```
that should definitely never be run with those arguments

### DIFF
--- a/fab/utils.py
+++ b/fab/utils.py
@@ -179,7 +179,7 @@ def traceback_string():
         trace=trace,
         type=exc_type.__name__,
         exc=exc,
-    ).replace('"', '\\"')
+    )
 
 
 def pip_install(cmd_prefix, requirements, timeout=None, quiet=False, proxy=None):
--- a/fab/utils.py
+++ b/fab/utils.py
@@ -179,7 +179,7 @@ def traceback_string():
         trace=trace,
         type=exc_type.__name__,
         exc=exc,
-    )
+    ).replace('"', '\\"')
 
 
 def pip_install(cmd_prefix, requirements, timeout=None, quiet=False, proxy=None):
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -29,6 +29,7 @@ import datetime
 import os
 import posixpath
 import yaml
+import pipes
 from distutils.util import strtobool
 
 from fabric import utils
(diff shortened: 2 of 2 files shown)


fab/utils.py
```diff
@@ -179,7 +179,7 @@ def traceback_string():
         trace=trace,
         type=exc_type.__name__,
         exc=exc,
-    )
+    ).replace('"', '\\"')
```
millerdev: It's not clear what you're escaping for here, so maybe this isn't the correct place to be doing it? It seems like you'd want to do this at the place where you're invoking the command. Also, would it make sense to use a function like [`shlex.quote()`](https://docs.python.org/3/library/shlex.html#shlex.quote), maybe [here](https://github.com/dimagi/commcare-hq-deploy/blob/96174966ec27b17f95e7f51e353d77abda9d8f11/fab/fabfile.py#L326-L328) on `subject`, `message`, and maybe `env.environment`?

benrudolph: all good points, will update

benrudolph:
updated, but ended up using `pipes` since `shlex.quote` is only for python3. realize it's deprecated (https://docs.python.org/2/library/pipes.html#pipes.quote) , but there didn't seem like a great alternative. i don't really think python3 is on the horizon so thought it was OK

millerdev:
Heh, didn't even realize `shlex.quote()` is only python 3

PR #159, opened 2016-12-16
track icds migration patches

sravfeyn:


fab/patches/2016-12-16_07.01.patch
```diff
@@ -0,0 +1,37 @@
+From ad056a9621f8782d31825b743457ef2da208629d Mon Sep 17 00:00:00 2001
```
czue: should this not also be in fab/patches/icds?

czue (commented):
thanks - one question

fab/patches/2016-12-16_07.01.patch
sravfeyn: sorry, updated to right directory

czue (approved):

PR #160, opened 2016-12-16
run git gc in current release

dannyroberts:
as part of clean_releases


benrudolph (approved):

dannyroberts:
in steady state, this will add about 15-20 seconds to deploy I think

benrudolph:
the only thing slightly weird about this is that it's slightly awkward function to put it in. clean_releases just deletes old repos, but don't feel that strongly about it

PR #163, opened 2016-12-27
Add formplayer to all source

benrudolph:
anyone

### DIFF
--- a/fab/const.py
+++ b/fab/const.py
@@ -9,6 +9,7 @@ ROLES_ALL_SRC = [
     'django_celery',
     'django_pillowtop',
     'formsplayer',
+    'formplayer',
     'staticfiles'
 ]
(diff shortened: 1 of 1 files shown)


fab/const.py
```diff
@@ -9,6 +9,7 @@
     'django_celery',
     'django_pillowtop',
     'formsplayer',
+    'formplayer',
```
czue, line 12: what's the difference between formsplayer and formplayer?

czue (approved):

czue:
assume it's correct to have both of those, though is weird. :+1: to merge from my side

fab/const.py
benrudolph: deprecated name for touchforms :/

czue: good times

PR #166, opened 2017-01-03
Patch for locationmanager

benrudolph:
checked out the commit before this was added and made my own commit to use a format-patch off of

@esoergel @czue


fab/patches/icds/2016-01-03-LocationManager-a.patch
```diff
@@ -0,0 +1,36 @@
...
++        locations and their children.
++        """
++        return SQLLocation.objects.get_queryset_descendants(
++            SQLLocation.objects.filter(location_id__in=location_ids),
++            include_self=True
++        )
++
++    def get_locations_and_children_ids(self, location_ids):
++        return list(self.get_locations_and_children(location_ids).location_ids())
++
```
esoergel, line 30: Looks good to me.

PR #169, opened 2017-01-10
Reverse apply patch

benrudolph:
anyone. used this to reverse my patches to NIC

cc: @gcapalbo @millerdev

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -427,7 +427,8 @@ def apply_patch(patchfile=None):
 
         fab <env> apply_patch:patchfile=/path/to/patch
 
-    Note: Only use this when absolutely necessary.
+    Note: Only use this when absolutely necessary. Normally we should use regular
+    deploy. This is only used for patching when we do not have access to the Internet.
     """
     if not patchfile:
         print red("Must specify patch filepath")
@@ -443,7 +444,8 @@ def reverse_patch(patchfile=None):
 
         fab <env> reverse_patch:patchfile=/path/to/patch
 
-    Note: Only use this when absolutely necessary.
+    Note: Only use this when absolutely necessary. Normally we should use regular
+    deploy. This is only used for patching when we do not have access to the Internet.
     """
     if not patchfile:
         print red("Must specify patch filepath")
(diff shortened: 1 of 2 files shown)


fab/operations/release.py
```diff
@@ -303,3 +303,16 @@ def apply_patch(filepath):
...
+@parallel
+def reverse_patch(filepath):
+    destination = '/home/{}/{}.patch'.format(env.user, env.deploy_metadata.timestamp)
+    operations.put(
+        filepath,
+        destination,
+    )
+
+    current_dir = sudo('readlink -f {}'.format(env.code_current))
+    sudo('git apply -R --unsafe-paths {} --directory={}'.format(destination, current_dir))
```
millerdev, line 318: This can fail, right? Do things go bad if that happens?

fab/fabfile.py
```diff
@@ -436,6 +436,22 @@ def apply_patch(patchfile=None):
...
 
 
+@task
+def reverse_patch(patchfile=None):
+    """
+    Used to reverse a git patch created via `git format-patch`. Usage:
+
+        fab <env> reverse_patch:patchfile=/path/to/patch
+
+    Note: Only use this when absolutely necessary.
```
millerdev: Might be worth expounding on why? I don't know the reasons, but I'm guessing because it might work and it might not, and if it fails it's probably going to be a bad time (possibly involving service downtime?).

benrudolph: Bad things don't actually happen the patch just doesn't get applied and the fab task stops. this should ~never be used, except that for the time being we can't deploy to NIC due to not being able to access the internet, so we're forced to use this method. can update the comment

millerdev: Haha, the things we do for NIC... That's good if it fails in a non-destructive way. I don't have all the flags to `git apply` in my head, but I was thinking that it could end up in a state where the patch was partially applied and needed human intervention—that probably involves a flag you're not using here.

benrudolph: Yeah NIC has caused so many headaches 😖 . For reference `-R` means reverse the specified patch, `--unsafe-paths` makes `git apply` a better gnu patch (instead of trying to create a git commit), and `--directory` is just the root dir from which the patch should be applied `code_current`. will elaborate in the docstring

millerdev (approved):
I'm happy to approve this as is if you either commit to fixing the docstring in another PR or think it's not even worth it (your judgement may be better than mine on that).

PR #170, opened 2017-01-10
add enikshay proxy

emord:
New proxy for uatbc @czue @proteusvacuum

### DIFF
--- a/fab/inventory/softlayer
+++ b/fab/inventory/softlayer
@@ -22,6 +22,9 @@
 [proxy1]
 10.162.36.208
 
+[proxy2]
+10.162.36.234
+
 [formplayer0]
 10.162.36.248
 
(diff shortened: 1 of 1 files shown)


proteusvacuum (approved):

emord:
This shouldn't block but staticfiles don't currently work on the enikshay endpoint. I think it's because staticfiles are only built for the main proxy https://github.com/dimagi/commcare-hq-deploy/blob/1207f183f127fa92dd87592ab034902f1e065a35/fab/fabfile.py#L284

Could start to be pretty painful to build staticfiles on 3 servers for every india deploy though, so I'm going to punt on changing that unless one of you thinks it would be necessary for https://enikshay.commcarehq.org to function as a full site

czue:
made this comment on the wrong PR, but yeah i don't think we need full site functionality.

PR #171, opened 2017-01-11
WIP Offline Deploy

benrudolph:
@snopoke @dannyroberts Here's an attempt at offline deploy. More of strategy than fully working code.

1. Run `fab prepare_offline_deploy`. This will clone HQ at latest master, depth 1. Bower install, and zip bower files. NPM install, and zip node_modules. `pip wheel` to download pip wheels, and zip those. All this is done in an `offline-staging` directory
2. Connect to VPN and run `fab <env> offline_deploy`.
3. This will clone the previous release to the user's home directory
4. Push the staging HQ directory to that directory (still need to figure out submodules)
5. Clone the user's release into the new `code_root`
6. upload and extract the bower/npm/pip wheels
7. Pip install from wheels.
8. Rest runs as normal with just a few minor tweaks
(still need to handle formplayer, but that should be easy)

I'm worried I'm going to get bogged down in some minor details that will make it not work. But thought I'd throw up what I had now and see what you guys thought

cc: @emord


fab/operations/offline.py
```diff
@@ -0,0 +1,54 @@
...
+)
+
+
+def prepare_zipfiles():
+    local(
+        'git clone --depth 1 --recursive https://github.com/dimagi/commcare-hq.git {}/commcare-hq'
+        .format(OFFLINE_STAGING_DIR)
+    )
+
+    # After we've created our HQ code zip, let's get bower and npm
```
emord: It doesn't look like you're creating the zip here. is that intentional?

If we only create at the end of this method, would it be possible to only transfer one zip file to the server?

fab/operations/offline.py
```diff
@@ -0,0 +1,54 @@
...
+    prepare_pip_wheels(os.path.join('requirements', 'requirements.txt'))
+    prepare_pip_wheels(os.path.join('requirements', 'prod_requirements.txt'))
+    zip_folder(
+        os.path.join(OFFLINE_STAGING_DIR, WHEELS_ZIP_NAME),
+        os.path.join(OFFLINE_STAGING_DIR, 'wheelhouse')
+    )
+
+
+def zip_folder(destination, folder):
+    local('zip -r {} {}'.format(destination, folder))
```
emord: Have you checked out using the different compression options here? If we have a lot of assets like jpgs in the repo or dependencies it might make sense to not compress at all if it saves a lot of time

fab/operations/release.py
```diff
@@ -66,6 +48,87 @@ def update_code(git_tag, use_current_release=False):
...
+
+def upload_pip_wheels():
+    _upload_and_extract(os.path.join(
+        OFFLINE_STAGING_DIR, WHEELS_ZIP_NAME
+    ))
+
+
+def _upload_and_extract(zipname):
+    put(zipname, env.offline_code_dir)
+    run('unzip {}/{}'.format(env.offline_code_dir, zipname))
```
emord: We don't have unzip on softlayer. I ran into that the other day uploading certs. Do we have it on NIC? May be better to use the python zip library, or tar.gz/tar.bz2

fab/operations/offline.py
emord: Also could look into using git archive or only copying the .git directory. I think you should be able to restore any commit from the .git directory, but copying the entire directory essentially copies everything twice

fab/fabfile.py
```diff
@@ -388,6 +390,21 @@ def deploy_formplayer():
 
 
 @task
+def offline_setup_release(keep_days=0):
```
emord: I may have missed it, but I think you also want to `copy_release_files` in here

emord (commented):
Looks like a great start, and seems like a good workflow

fab/operations/offline.py
benrudolph: Yeah out dated comment. realized it's hundreds of megabytes even after zipping and using depth = 1. seems prohibitively large to try and upload. i like the idea of just copying the `.git`. will give that a go

fab/operations/offline.py
benrudolph: nope haven't looked into the different types, sounds like a good optimization though

fab/operations/release.py
benrudolph: blergh 😞

fab/fabfile.py
benrudolph: ah yes good catch. need to do that

fab/operations/offline.py
czue: I thought i made this comment already but don't see it here. Another possibility that might be easier is to just setup a git server on one of the machines and push/pull to it from behind the VPN. I've done this on some personal projects and it's not very hard (though might be hard if we want to do it via ansbile... not sure if we feel that would be necessary)

benrudolph: i talked about this a little with simon. i forgot the reasons why i convinced myself that was hard, maybe that was before i realized we could still git push to a remote repo on the NIC cluster. one difficulty would just be setting up that git server without internet access. couldn't do it with ansible, but perhaps it's ok to just hack something up

czue: i've use [gitosis](https://git-scm.com/book/no-nb/v1/Git-on-the-Server-Gitosis) which looks like it just installs from a git repo which should be doable. only providing this information if useful, not advocating for changing what you have here since i'm sure you thought about this more than me

benrudolph: no this is definitely a step which i'm least confident in. it's very intensive process to attempt to clone the whole repo and then upload and/or push to each server. open to new ideas

emord: I like that idea, but I'm not sure how easy that would be with submodules. https://github.com/dimagi/commcare-hq/blob/master/.gitmodules has the github url. and I'm not sure how easy it would be for that to sometimes be github.com and sometimes some private server. Could have some entry for github in /etc/hosts, but that doesn't seem like a good solution to me

benrudolph: we've actually already solved that problem when we do our initial clone from previous release. essentially you can override the url by passing in the url directly to the clone command

benrudolph:
alrighty @czue this should ~about ready. i'm trying to get through the entire deploy my self, but here's what you'd need to run. i've gotten through a full offline_setup_release which I think is the hardest part, but there may be smaller things down the line that i haven't anticipated

```
fab prepare_offline_deploy
fab <env> deploy:offline=yes
```

emord (approved):
Looks good here. Small nit that you used zip in variable/function names even though it's been changed to tar.gz

czue:
@benrudolph do you have concerns about me running that directly on the server? how sure are we that we aren't going to bork something and get into a hard-to-recover-from state?

benrudolph:
I think we could always rollback in a pinch, but definitely don't feel
comfortable running on anything other than staging at this point. Sorry
should've specified that

benrudolph:
Though I suppose it no longer makes sense for you to try if it's for
staging haha

edit: new idea. i'm going to test from ansible to staging so i can actually make some progress. then when we're confident we can do it on icds

benrudolph:
ok quite a bit of churn at the end, but i've verified a working offline deploy.

on staging i was able to get it to properly pip install requirements at a different version, update submodules to different commits, and unpack static files correctly.

i also tested regular deploy which worked fine after the offline deploy. i think we're about ready to go for icds

fab/operations/release.py
```diff
@@ -143,10 +144,14 @@ def offline_pip_install():
         env.sudo_user, env.virtualenv_root
     )
     requirements = os.path.join(env.code_root, 'requirements')
+    requirements_file = os.path.join(requirements, 'requirements.txt')
+    for offline_lib in ['pygooglechart', 'django-transfer', 'django-two-factor-auth', 'pyzxcvbn']:
```
benrudolph, line 149: 👍 frustrating, but seems like a good work around. gonna be fun when someone upgrades one of these...

benrudolph:
think we should merge this? i think it does complicate the deploy a bit, but there are some benefits to having this in our back pocket. i also think it will speed up deploy time because this exercise taught me that the `git -c` option doesn't work with the submodule urls and that we are most likely cloning each submodule from scratch

czue:
i haven't been reviewing code closely (happy to though) but I do think this should be a tool in our toolkit at least until the situation gets easier so I'm in favor of getting this merged to master once there aren't issues

benrudolph:
@czue do you still need https://github.com/dimagi/commcare-hq-deploy/pull/171/commits/69382af69f0c969538039e5652cb5727251f0d58 after we stopped checking out master?

czue:
i think i'll be able to tell you on the next deploy but I think so. i needed it for this one even after i thought i stopped checking out master everywhere. My sense is that git repos in general are designed to never be pushed to if they've been created by a clone. not sure if we could update [these lines](https://github.com/dimagi/commcare-hq-deploy/pull/171/files#diff-265182d2e8204f0a07a2f9dcd2321ea7R207) to do a repo init / push workflow instead?

(i could be totally wrong about this though)

czue:
actually it looks like maybe if we use `git clone --bare` there it might work? http://stackoverflow.com/a/738256/8207

benrudolph:
ahh i think bare just might work:

` current -> clone  with --bare to home dir --> clone without bare to code_root`

wonder if this'll cause headaches with submodules

czue:
i was getting an error when i tried it.

fab/operations/release.py
emord: Is this just because the requirements of these are listed as either git or https? If so, we might be able to split those out to a separate reqs file

benrudolph: yes it is, that's a good idea too

benrudolph:
@snopoke https://github.com/dimagi/commcare-hq-deploy/pull/171/commits/b971284d277fc3cc89f39cdc06d937928d274710 this'll create a tag for offline deploys. i don't use github3py because that creates the tags on the remote repo which is not actually what we're deploying. we're deploying our local clone

snopoke (approved):

snopoke:
nice, thanks

benrudolph:
approved, going ahead and merging. new git is on pillow0 so we should be good to use this

czue:
amazing, thanks for the persistence on this! very glad we did it.

PR #176, opened 2017-01-27
Add formplayer to services

benrudolph:
@czue @wpride this is why softlayer seeming never got updates. supervisor never restarted...

cc: @emord @snopoke

### DIFF
--- a/fab/const.py
+++ b/fab/const.py
@@ -18,6 +18,7 @@ ROLES_ALL_SERVICES = [
     'django_celery',
     'django_pillowtop',
     'formsplayer',
+    'formplayer',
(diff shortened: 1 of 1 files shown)


fab/const.py
```diff
@@ -18,6 +18,7 @@
     'django_celery',
     'django_pillowtop',
     'formsplayer',
+    'formplayer',
```
snopoke, line 21: `formsplayer` & `formplayer` that's just confusing, no wonder you didn't notice it before

benrudolph: counting the days until we can get rid of touchforms...

PR #178, opened 2017-02-02
Increase max perm gen size

benrudolph:
@wpride @snopoke this increases the max perm gen size on formplayer. i inspected the jmx memoryusage bean and found that we were are near the max on that:

```
Usage = {
  committed = 85983232;
  init = 22020096;
  max = 85983232;
  used = 85706448;
 };
```
this memory is used for storing permanent objects like class meta data. trying to get this info into datadog. there are some good posts about all the different memory types in java, here's one that's a quick read: http://www.freshblurbs.com/blog/2005/05/19/explaining-java-lang-outofmemoryerror-permgen-space.html
cc: @emord


snopoke:
I've spent countless hours researching and messing with Java memory and GC settings!

benrudolph:
It's crazy they've pushed this down to the user level!

PR #181, opened 2017-02-17
standard deploy tag

snopoke:


fab/utils.py
```diff
@@ -90,34 +91,31 @@ def tag_commit(self):
...
                 staging_dir=OFFLINE_STAGING_DIR,
-                tag=tag_name,
+                tag=self._deploy_tag,
             ))
+        self._tag_created = True
 
     @property
     def diff_url(self):
-        if env.offline:
-            return '"No diff url for offline deploy"'
```
benrudolph, line 118: This gets tricky when you remove this, i don't think `self._last_tag` exists for an offline deploy since it searches for the last tag on github. however this could just be a remnant of me not being able to access outside internet from my local machine while on the vpn... so i think probably fine to keep

benrudolph (approved):
👍 let me know if this has been tested or not

snopoke:
I haven't tested this. I assumed that if we were able to create a tag then
we'd be able to search for the last tag as well.

benrudolph:
👍 you're right, makes sense

PR #182, opened 2017-02-21
Add 1G swap file to hqdb1-staging

kaapstorm:
@snopoke is this what you had in mind?

Does this need to be merged in conjunction with [commcarehq-ansibe #609](https://github.com/dimagi/commcarehq-ansible/pull/609), or will it work on its own? The only other example I could find of this was for riakcs, and riakcs has the `swap` role associated with it.

@benrudolph cc @wspride

### DIFF
--- a/fab/inventory/staging
+++ b/fab/inventory/staging
@@ -6,7 +6,7 @@ hqdjango0-staging.internal-va.commcarehq.org
 hqdjango1-staging.internal-va.commcarehq.org
 
 [postgresql]
-hqdb1-staging.internal-va.commcarehq.org
+hqdb1-staging.internal-va.commcarehq.org swap_size=1G
 
 [rabbitmq]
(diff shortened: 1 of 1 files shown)


snopoke:
Yes, that's what I was talking about. I'd want to be sure though that this won't create swap for our prod machines.

benrudolph:
updated this to be more machines (since others were running out of space). ran this on staging with `run_role` playbook. don't think we have to worry about this being added to prod servers because the `swap` role is not part of out common deploy: https://github.com/dimagi/commcarehq-ansible/search?utf8=%E2%9C%93&q=swap. even if it was i think the role is smart enough to not add swap if the inventory var doesn't exist

dannyroberts:
This is super old, but looks like we already deployed these changes and forgot to merge. I'm just going to merge because it seems harmless

PR #191, opened 2017-03-01
Hqcelery2

javierwilson:
Moving queues to the new hqcelery2 @benrudolph @gcapalbo


gcapalbo:
:+1: nice, thanks @javierwilson, the split looks good

PR #193, opened 2017-03-03
increase repeat record queue concurrency

dannyroberts:


emord (approved):

fab/environments.yml
```diff
@@ -44,7 +44,7 @@ softlayer:
       email_queue:
         concurrency: 2
       repeat_record_queue:
-        concurrency: 3
+        concurrency: 10
```
gcapalbo: i think we can make softlayer be at least what prod used to be (50)

fab/environments.yml
```diff
@@ -100,7 +100,7 @@ production:
       email_queue:
         concurrency: 2
       repeat_record_queue:
-        concurrency: 50
+        concurrency: 100
```
gcapalbo, line 103: as we chatted about offline, if this doesn't help we'll need to just add another worker

PR #196, opened 2017-03-08
new pillow hqpillowtop1

javierwilson:
@snopoke @benrudolph


benrudolph (approved):

benrudolph:
This will be picked up in next deploy. Do we want that? Will this mean both
machines will be working off the same offsets, and does that mean we'll
reactivate the couch to Kafka pillow on pillow1?

snopoke:
Yea, I don't think we should deploy this yet @javierwilson can you revert please

PR #203, opened 2017-03-13
Blacklist touchforms and add back formplayer

benrudolph:
@wpride this is the last hold out of old touchforms. once i get back to the office i'm going to deploy localsettings and then deploy this

cc: @biyeun


benrudolph:
Ok this is g2g, settings are up to date. i decided to keep around the touchforms process to make it easier to transition over

benrudolph:
ah shoot i thought this had been approved

PR #205, opened 2017-03-13
Add ucr indicator queue

emord:
@gcapalbo @benrudolph

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -43,6 +43,8 @@ softlayer:
         concurrency: 8
       ucr_queue:
         concurrency: 4
+      ucr_indicator_queue:
+        concurrency: 10
       email_queue:
         concurrency: 2
       repeat_record_queue:
@@ -80,6 +82,8 @@ icds:
         concurrency: 8
       ucr_queue:
         concurrency: 4
+      ucr_indicator_queue:
+        concurrency: 4
       email_queue:
         concurrency: 2
       repeat_record_queue:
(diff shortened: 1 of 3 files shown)


fab/environments.yml
```diff
@@ -107,6 +111,8 @@ production:
         concurrency: 200
       ucr_queue:
         concurrency: 4
+      ucr_indicator_queue:
+        concurrency: 10
```
gcapalbo, line 116: is 10 needed? that might be a lot

also i think it would be good to put this on celery2

gcapalbo: oh nevermind, i see you're using gevent, this is fine as is

gcapalbo (commented):
we should add this to staging and swiss too right?

gcapalbo:
looks good. before releasing this it would be good to test it either locally or on staging to make sure it works as intended. i remember @benrudolph had to add some [patching](https://github.com/dimagi/commcare-hq/pull/10824) to manage.py to get celery working with gevent, so we should just make sure everything still works as intended with the patching

emord:
Swiss doesn't have UCRs https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/pillows/swiss.yml#L8

I wasn't going to add it to staging since we constantly have issues with it and didn't want to increase those problems. I will double check this works locally tomorrow morning, but ben's  changes look like it applies to all celery commands

benrudolph:
confirming that that patching should work for all queues in the future

gcapalbo:
yup, agreed we shouldn't have to make changes to the patching - i just meant that the patching creates an environment under which we haven't run these tasks yet so we should run them in that environment to make sure it still does what we want

PR #206, opened 2017-03-14
add intrahealth pillow

emord:
@calellowitz (i think you'll be deploying today) Let me know if you have any issues with deploying this.

buddy @gcapalbo

### DIFF
--- a/fab/pillows/production.yml
+++ b/fab/pillows/production.yml
@@ -1,19 +1,10 @@
 # fab/pillows/production.yml
 ---
-exclude_groups:
-  - mvp_indicators
-# include_groups:
-#   - mvp_indicators
 include_pillows:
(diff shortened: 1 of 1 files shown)


fab/pillows/production.yml
```diff
@@ -1,19 +1,10 @@
 # fab/pillows/production.yml
 ---
-exclude_groups:
-  - mvp_indicators
```
emord, line 4: Looks like @krsdimagi said we're only supporting their reports up until 2/28 so not going to add this pillow back in. Going to leave deleting mvp reporting code for someone who will find it more cathartic

gcapalbo (approved):

emord:
This checkpoint will need to be set to 49849574 (min of the current pillows) i can handle that after deploy

PR #207, opened 2017-03-16
add split pillows to prod

emord:
@snopoke good idea https://github.com/dimagi/commcare-hq/pull/15351#issuecomment-286987408

@gcapalbo


snopoke (changes requested):
Can you add the split pillows to ICDS as well

emord:
done

snopoke (approved):

PR #210, opened 2017-03-21
further split ICDS static UCR pillows

snopoke:
@emord @benrudolph

### DIFF
--- a/fab/pillows/icds.yml
+++ b/fab/pillows/icds.yml
@@ -2,5 +2,7 @@ exclude_pillows:
   - ReportCaseToElasticsearchPillow
   - ReportXFormToElasticsearchPillow
 include_pillows:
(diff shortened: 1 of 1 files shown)


emord:
@snopoke can you add these to default.yml?

I think you also need to add these in settings.py on HQ

snopoke:
I don't think they need to be in `default.yml` since they're only going to get added to ICDS localsettings.

Also they shouldn't be in `settings.py` since they're only for ICDS - unless you're saying I should just add them there rather than into ICDS localsettings.

emord:
Ah. ok. I didn' realize you're only adding it to the localsettings.

emord (approved):

benrudolph:
very cool

PR #212, opened 2017-03-21
make offline deploy faster

snopoke:
decided to just make some of these changes since I wanted to do a deploy anyways:

* use ansible user
* re-use remote offline code root
* rsync files to one machine from local and then rsync from that machine to all the others

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -863,5 +863,4 @@ OFFLINE_DEPLOY_COMMANDS = [
     db.flip_es_aliases,
     staticfiles.update_manifest,
     release.clean_releases,
-    release.clean_offline_releases,
 ]
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -191,7 +191,7 @@ def icds():
     env.force = True  # don't worry about kafka checkpoints on icds
     # Force ansible user and prompt for password
     env.user = 'ansible'
-    env.sudo_password = getpass('Enter the password for then ansbile user: ')
+    env.password = getpass('Enter the password for then ansbile user: ')
     execute(env_common)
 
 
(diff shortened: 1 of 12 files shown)


calellowitz (approved):
👍 this looks great and will be super useful if we get stuck doing more offline deploys.
Thanks for doing this.

fab/fabfile.py
```diff
@@ -186,6 +189,9 @@ def icds():
     env.inventory = os.path.join(PROJECT_ROOT, 'inventory', 'icds')
     load_env('icds')
     env.force = True  # don't worry about kafka checkpoints on icds
+    # Force ansible user and prompt for password
+    env.user = 'ansible'
+    env.sudo_password = getpass('Enter the password for then ansbile user: ')
```
benrudolph: should we only do this if we're doing an offline deploy?

fab/operations/release.py
```diff
@@ -58,6 +55,44 @@ def create_offline_dir():
...
+            os.path.join(OFFLINE_STAGING_DIR, 'commcare-hq', sync_item),
+            delete=True,
+        )
+    rsync_project(
+        env.offline_code_dir,
+        os.path.join(OFFLINE_STAGING_DIR, 'formplayer.jar'),
+    )
+
+
+def sync_offline_from_control():
```
benrudolph, line 77: sweet

benrudolph (approved):
nicely done 👏

benrudolph:
@snopoke just an fyi adding the control machine to the inventory files causes ansible to fail on the become user. it doesn't stop us from executing tasks though
```
fatal: [control.internal-va.commcarehq.org] => Incorrect become password
```

snopoke:
is that because the control machine doesn't have the `cchq` user?

benrudolph:
i believe so

sravfeyn:
Hey @snopoke should we revert some of these changes, now that we don't need them anymore? For e.g. ansible password prompt when doing icds deploy

benrudolph:
i think we should skip the prompt and ansible user if the deploy isn't offline. otherwise i think these changes are good, you never know when we may have to deploy offline again...

sravfeyn:
Yeah definitely, meant to revert only any hacks like the example I gave.

PR #215, opened 2017-03-24
adds patch server and patch clients

javierwilson:
ads options for patch server and patch clients to icds inventory


fab/inventory/icds
```diff
@@ -117,5 +120,20 @@ kafka0
...
+pg0
+pg1
+web0
+web1
+es0
+es1
+couch0
+celery0
+pil0
+kafka0
```
snopoke, line 136: is this 'all server except patch0'? If so I think we could generate the list dynamically. Maybe not worth it if it's hard though.

PR #216, opened 2017-03-24
Increase concurrency on background worker

gcapalbo:
Memory usage is good on the machine, seems fine to do

@proteusvacuum


gcapalbo:
closing this for now, might have been a coincidence but this may have put a lot more load on es when i boosted it manually in flower

PR #219, opened 2017-03-25
GEvent for background queue

benrudolph:
@gcapalbo going to put this up on staging

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -27,7 +27,7 @@ softlayer:
       main:
         concurrency: 8
       background_queue:
-        concurrency: 4
+        concurrency: 20
       periodic:
         concurrency: 4
         server_whitelist: 10.162.36.233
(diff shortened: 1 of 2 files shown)


gcapalbo (approved):
sounds good, i think one thing to be aware of is that i had tried boosting the concurrency only to 8 through flower on friday and we got a lot of es errors when i did that. might be less of an issue now though since there should be less contention with rabbit not being hosted there

gcapalbo:
@benrudolph i merged master to resolve conflicts with staging, and also removed the max tasks per child setting since it doesn't apply with gevent

i'm also starting to think concurrency 20 might be too much with the other issues we've been having

benrudolph:
thanks this kind of fell to the backlog. agree that the high concurrency is probably not the wisest thing to do given the recent issues

nickpell:
conflicting with master - is it safe to remove this branch from staging.yaml?

benrudolph:
Yea sorry

PR #221, opened 2017-03-25
adds real proxy address (patch)

javierwilson:
otherwise it tries to use the softlayer's proxy

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -58,7 +58,7 @@ softlayer:
 
 icds:
   environment: icds
-  http_proxy: 169.38.84.79:3128
+  http_proxy: 10.247.63.132:3128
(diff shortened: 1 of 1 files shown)


fab/environments.yml
```diff
@@ -58,7 +58,7 @@ softlayer:
 
 icds:
   environment: icds
-  http_proxy: 169.38.84.79:3128
+  http_proxy: 10.247.63.132:3128
```
sravfeyn, line 61: Is it not easy to add this to inventory file, so that ansible and fab can pull from same place?

snopoke: might be an option but going to merge this for now

snopoke (approved):

PR #226, opened 2017-03-31
fixing monolith check

calellowitz:
@snopoke @benrudolph 
Since adding the control server to the inventory, swiss was not being registered as a monolith and it broke the deploy process


benrudolph (approved):
good catch, thanks

snopoke:
Thanks

snopoke:
I'd like to say lets remove it but we may need offline deploy again.

PR #228, opened 2017-04-04
Revert "add proxy vars to supervisor config"

calellowitz:
Reverts dimagi/commcare-hq-deploy#224
@emord cc: @snopoke @benrudolph @javierwilson 
the proxy variable was causing the webworker requests to riak (and i think formplayer) to be routed through the squid proxy and then fail. we fixed this manually on NIC but this will help future deploys until we figure out the proper way to do add in the proxy.

### DIFF
--- a/fab/operations/supervisor.py
+++ b/fab/operations/supervisor.py
@@ -244,18 +244,15 @@ def _format_env(current_env, extra=None):
     newrelic_machines = [machine.name
                          for group in inventory_groups for machine in group.hosts
                          if 'newrelic_app_name' in group.vars]
-
-    ret['new_relic_command'] = ''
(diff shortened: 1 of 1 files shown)


benrudolph:
thanks for handling 💪 sounds like we need a way to specify which supervisor templates get these...

snopoke:
😐 thanks @calellowitz. The main use case for this was sentry.

benrudolph:
i think we can get around this with `NO_PROXY=machine1,machine2,...`. shouldn't be too hard to add

javierwilson:
I've added https://github.com/dimagi/commcare-hq-deploy/pull/229

PR #229, opened 2017-04-04
adds no_proxy

javierwilson:
whoever honors http_proxy *should* also honor no_proxy.
adding no_proxy for all hosts in inventory.

### DIFF
--- a/fab/operations/supervisor.py
+++ b/fab/operations/supervisor.py
@@ -253,9 +253,12 @@ def _format_env(current_env, extra=None):
         ret['supervisor_env_vars']['NEW_RELIC_CONFIG_FILE'] = '%(root)s/newrelic.ini' % env
         ret['supervisor_env_vars']['NEW_RELIC_ENVIRONMENT'] = '%(environment)s' % env
 
+    all_hosts = [host.name for host in InventoryParser(env.inventory).groups['all'].get_hosts()]
+
     if env.http_proxy:
         ret['supervisor_env_vars']['http_proxy'] = 'http://{}'.format(env.http_proxy)
         ret['supervisor_env_vars']['https_proxy'] = 'https://{}'.format(env.http_proxy)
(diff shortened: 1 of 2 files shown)


emord (approved):
thanks @javierwilson

fab/operations/supervisor.py
```diff
@@ -253,9 +253,12 @@ def _format_env(current_env, extra=None):
         ret['supervisor_env_vars']['NEW_RELIC_CONFIG_FILE'] = '%(root)s/newrelic.ini' % env
         ret['supervisor_env_vars']['NEW_RELIC_ENVIRONMENT'] = '%(environment)s' % env
 
+    all_hosts = [host.name for host in InventoryParser(env.inventory).groups['all'].get_hosts()]
+
     if env.http_proxy:
         ret['supervisor_env_vars']['http_proxy'] = 'http://{}'.format(env.http_proxy)
         ret['supervisor_env_vars']['https_proxy'] = 'https://{}'.format(env.http_proxy)
+        ret['supervisor_env_vars']['no_proxy'] = '{},relay.nic.in,smsgw.sms.gov.in,icds-cas.gov.in'.format(','.join(all_hosts))
```
benrudolph: do you think we can put these extra hosts in a variable in `environments.yml`? something like `additional_no_proxy_hosts` maybe?

benrudolph (commented):
nice, one small comment. thanks for the quick turnaround

fab/operations/supervisor.py
javierwilson: so taht they only affect the specific environment? that would make sense

javierwilson:
it now uses additional_no_proxy_hosts from environment.yml

emord (approved):
Looks good to me @javierwilson is there any reason to not merge this? it still has the don't merge label on

benrudolph (approved):

PR #230, opened 2017-04-06
Reduce max tasks per child

gcapalbo:
@calellowitz @benrudolph 

In the short term this might help a little bit with the out of memory errors (I think the longer term solution is to do some profiling on the tasks running on the machine. There seemed to be a lot of `corehq.apps.export.tasks.rebuild_export_task` tasks running when it happens which might be related.)

Essentially with this we'll replace a worker child process after every task instead of every 5 tasks, so for example in the background worker since it has concurrency 4, instead of keeping in memory the effects of up to 20 tasks, it will only keep around the effects of up to 4 tasks.


benrudolph (approved):
Thanks gio that makes sense and a good short term work around. i did a lot digging into the rebuild export task last night and nothing is jumping out as hugely memory intensive. one thing to note is that when we do our ES scroll query, we grab 500 forms at one time and load them into memory. that can potentially add up quickly to a lot of forms in memory, but still seems weird to me that it would cause such a memory bottleneck. working on adding some logging

PR #232, opened 2017-04-07
NIC 60k upgrade inventory

sravfeyn:
@javierwilson @calellowitz @emord @snopoke 

Don't merge yet

### DIFF
--- a/fab/inventory/icds
+++ b/fab/inventory/icds
@@ -39,6 +39,37 @@
 [control]
 10.247.24.20
 
+[pg2]
+10.247.24.22 devices='["/dev/sdb"]' partitions='["/dev/sdb1"]' datavol_device='/dev/mapper/consolidated-data' hot_standby_server='10.247.24.28'
+
+[pg3]
+10.247.24.23 devices='["/dev/sdb"]' partitions='["/dev/sdb1"]' datavol_device='/dev/mapper/consolidated-data' hot_standby_server='10.247.24.29'
+
+[web2]
+10.247.24.24 devices='["/dev/sdb"]' partitions='["/dev/sdb1"]' datavol_device='/dev/mapper/consolidated-data'
+
+[web3]
+10.247.24.25 devices='["/dev/sdb"]' partitions='["/dev/sdb1"]' datavol_device='/dev/mapper/consolidated-data'
+
+[web4]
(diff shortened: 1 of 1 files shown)


fab/inventory/icds
```diff
@@ -72,28 +109,35 @@ kafka0
 
 [celery:children]
 celery0
+celery1
 
 [pillowtop:children]
 pil0
```
sravfeyn: We will need to add `pil1` here after we update HQ to support multiple pillow machines. We will also need to do same for kafka and add `pil1` to `kafka:children`

fab/inventory/icds
```diff
@@ -72,28 +109,35 @@ kafka0
...
 
 [touchforms:children]
-pil0
+pil1
```
sravfeyn, line 120: Should we take any special care when moving touchforms over while doing ansible deploy?

fab/inventory/icds
```diff
@@ -39,6 +39,37 @@
 [control]
 10.247.24.20
 
+[pg2]
+10.247.24.22 devices='["/dev/sdb"]' partitions='["/dev/sdb1"]' datavol_device='/dev/mapper/consolidated-data' hot_standby_server='10.247.24.28'
```
sravfeyn, line 43: - [ ] To update actual devices after partition or don't need it if doing without extra storage

fab/inventory/icds
```diff
@@ -48,11 +79,17 @@ web0
...
+web3
+web4
+web5
 kafka0
 
 [postgresql:children]
 pg0
 pg1
+pg2
+pg3
```
sravfeyn, line 92: How should the shards be updated [here](https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/vars/icds/icds_public.yml#L81)?

fab/inventory/icds
benrudolph: touchforms is stateless so it shouldn't matter

emord:
I think you guys will also want to update the celery queues https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/environments.yml#L67

emord:
@sravfeyn can this be closed out? I think all of these have been deployed at some point

dannyroberts:
closing

PR #234, opened 2017-04-16
65k django workers icds

sravfeyn:


sravfeyn:
This is already ansible deployed and tested

PR #237, opened 2017-04-21
new machines

sravfeyn:
Don't merge to master yet. Going to deploy with this inventory. Will ping back for merge once migration etc is over

PR #241, opened 2017-04-21
Only deploy tasks to celery0

emord:
Don't really need to split these up right now, but realized this is necessary when https://github.com/dimagi/commcare-hq-deploy/pull/237 is deployed. otherwise it would have run two instances of the queues, one on each machine. I wasn't sure if that was a safe operation or not.

@gcapalbo @snopoke @calellowitz

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -78,7 +78,6 @@ icds:
         concurrency: 4
       ucr_indicator_queue:
         concurrency: 4
-    'celery1':
       background_queue:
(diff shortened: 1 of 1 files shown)


gcapalbo (approved):

gcapalbo:
as a side note - it's completely fine to have more than one worker running consuming from the same queue, and is something we will likely have to do some day. it's jut important that each worker has a different hostname param when starting the worker up

PR #242, opened 2017-04-22
support multiple pillow machines

emord:
Copying the format of celery processes, it's repetitive right now, but eventually that will change when they have differen processes

@dimagi/scale-team @millerdev


fab/services/templates/supervisor_pillowtop.conf
```diff
@@ -1,5 +1,5 @@
 [program:{{ project }}-{{ environment }}-pillowtop-{{ pillow_name }}]
-command={{ virtualenv_current }}/bin/python {{ code_current }}/manage.py run_ptop {{ pillow_option }}
+command={{ virtualenv_current }}/bin/python {{ code_current }}/manage.py run_ptop {{ pillow_option }} --num_processes={{ num_processess }} --process_num={{ process_num }}
```
snopoke, line 2: where to these vars come from?

snopoke: never mind - I see it in the other PR

snopoke (approved):

snopoke:
Might be good to add a line to docs somewhere explaining how to add / remove a pillow.

millerdev (approved):
Kinda hard to verify that the old `fab/pillows/*.yml` contents got transferred into new pillow configs in `fab/environments.yml`—assuming @emord has double-checked that.

emord:
@millerdev yeah. that was the point of making htem alphabetical. when I went back through and checked I had missed a pillow.

@snopoke will add some docs

fab/environments.yml
```diff
@@ -54,6 +54,58 @@ softlayer:
...
+      UnknownUsersPillow:
+        num_processes: 1
+      UserCacheInvalidatePillow:
+        num_processes: 1
+      UserGroupsDbKafkaPillow:
+        num_processes: 1
+      UserPillow:
+        num_processes: 1
+      XFormToElasticsearchPillow:
+        num_processes: 1
```
dannyroberts, line 108: Maybe this is fine (not sure), but I'm wondering, does this mean that when a new pillow is added, it won't go into production (or softlayer, or icds) until it's also added here? A nice thing about the exclude/include stuff is that it made the default clear and small deviations from the default easy and clear in how they differed from it. (I get that that wasn't easily extensible to specifying which pillows go on which machines, though.)

dannyroberts (approved):
Didn't review in great detail, but the approach seems reasonable. Mostly I love that we're finally getting to this

fab/environments.yml
emord: Yeah. I had the same conflicted feelings about that, but need to get this out before pillow partitioning. I would be pro a refactor for both celery and pillows that enables including and excluding, but I think it's good that we're using the same type of config now

dannyroberts: :+1:

PR #243, opened 2017-04-26
don't use quotes

emord:
@sravfeyn @calellowitz I think this is why there weren't supervisor configs in the new directory

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -64,7 +64,7 @@ icds:
   formplayer_memory: "16000m"
   gunicorn_workers_static_factor: 1
   celery_processes:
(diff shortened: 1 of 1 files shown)


emord:
Actually this doesn't work. something doesn't work with our deploy right now. celery isn't getting any celery supervisor tasks. set to it.  For now i copied the services from the previous release into current/services and I can take a look at this tomorrow, unless someone else wants to do it

benrudolph:
i've seen that before when the management command to generate those templates silently failed

emord:
This is safe to merge. I'll debug more if it fails again today

PR #253, opened 2017-05-10
adds hostnames in vars for icds

javierwilson:


snopoke:
How do these get used?

javierwilson:
the hostname is defined for the system (using hostnamectl, i think) and also changes /etc/hosts, this way datadog and others register the host as "web5" instead of whatever NIC assigns, it also allows us to use a common domain name that matches DNS records, instead of using "gov.in" as it is by default.

PR #256, opened 2017-05-11
bump up ucr workers

emord:
@dannyroberts as mentioned https://github.com/dimagi/commcare-hq/pull/16122

celery1 is at about 50% cpu. celery0 is < 10% so this shouldn't cause any resource issues

@gcapalbo you said that it's acceptable to have the same worker on two different machines because they have different hostnames right?

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -144,10 +144,12 @@ icds:
         concurrency: 2
       async_restore_queue:
         concurrency: 4
+      ucr_indicator_queue:
+        concurrency: 8
       flower: {}
     '10.247.24.31': # celery1
       ucr_indicator_queue:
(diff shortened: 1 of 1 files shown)


snopoke (approved):
:+1: assuming @gcapalbo confirms your question.

gcapalbo:
yeah it's fine to have workers on different (or even the same) machine consuming from the same queue.

the only thing to make sure of is that they get a different --hostname param when starting the worker up: https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/services/templates/supervisor_celery_ucr_queue.conf#L3

right now this should happen automatically when they're on different machines based on how we populate that param

PR #261, opened 2017-05-23
inventory and environment for l10k

javierwilson:


fab/environments.yml
```diff
@@ -477,3 +477,67 @@ swiss:
...
+        concurrency: 1
+      email_queue:
+        concurrency: 1
+      saved_exports_queue:
+        concurrency: 1
+      pillow_retry_queue:
+        concurrency: 1
+      ucr_queue:
+        concurrency: 1
+      async_restore_queue:
```
benrudolph, line 600: wonder if we can get away with only a couple queues. we probably don't need this, but suppose it doesn't make too much of a difference now that we use autoscale

fab/inventory/l10k
```diff
@@ -0,0 +1,61 @@
+[l10kserver1]
+10.162.36.254
+
+#[l10kbackup1]
+#ip_backup_server
```
benrudolph, line 5: one thing i was thinking about for the backup server is that i'm not sure if we'll have an internal vpn for the two servers. we breaks a lot of the assumptions in the ansible code. for example if we streamed postgres data to the backup server, it'd be in plain text. something we should probably think about/discuss

benrudolph (commented):
sweet, assuming we'll leave this un-merged until we get the actual server?

fab/inventory/l10k
javierwilson: true. but also, even if there no VPN, if both servers are plugged into the same switch, and the both have two NICs (as servers usually do) we can use one of the NICs for 10.0.0.x and use it as a LAN, also if in the same switch, the switch should only transfer from one machine to the other, so same security as VPN.  Still, I'd prefer to have the both in a LAN and have different firewall rules for LAN and WAN.

javierwilson:
and yes i think we should leave this un-merged until we get the actual server, that's the plan.

fab/inventory/l10k
benrudolph: that's good to know thanks for that explanation

fab/environments.yml
```diff
@@ -574,3 +574,67 @@ swiss:
...
+        num_processes: 1
+      GroupPillow:
+        num_processes: 1
+      GroupToUserPillow:
+        num_processes: 1
+      KafkaDomainPillow:
+        num_processes: 1
+      SqlSMSPillow:
+        num_processes: 1
+      UnknownUsersPillow:
```
snopoke, line 631: can remove this one

fab/inventory/l10k
```diff
@@ -0,0 +1,64 @@
...
+l10kserver1
+
+[shared_dir_host:children]
+l10kserver1
+
+#[pg_standby:children]
+#l10kbackup1
+
+#[lvm:children]
+#l10kserver1
```
snopoke, line 61: can just remove this completely

snopoke (commented):
Going to merge despite comments since none of them are blocking.

PR #275, opened 2017-06-07
scale out ICDS pillows

snopoke:
- [x] added partitions to kafka for `case-sql` (8), `form-sql` (8) and `sms` (4)
- [x] created new checkpoints with `./manage.py add_kafka_partitions`

Increases based off this data:

image

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -163,8 +163,6 @@ icds:
         num_processes: 1
       CacheInvalidatePillow:
         num_processes: 1
-      CaseSearchToElasticsearchPillow:
-        num_processes: 1
       DefaultChangeFeedPillow:
         num_processes: 1
       DomainDbKafkaPillow:
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -166,39 +166,39 @@ icds:
       CaseSearchToElasticsearchPillow:
         num_processes: 1
       CaseToElasticsearchPillow:
-        num_processes: 1
+        num_processes: 4
       DefaultChangeFeedPillow:
         num_processes: 1
       DomainDbKafkaPillow:
         num_processes: 1
       FormSubmissionMetadataTrackerPillow:
-        num_processes: 1
+        num_processes: 2
       GroupPillow:
         num_processes: 1
       GroupToUserPillow:
         num_processes: 1
       kafka-ucr-main:
-        num_processes: 1
+        num_processes: 2
       kafka-ucr-static-awc-location:
-        num_processes: 1
+        num_processes: 4
       kafka-ucr-static-cases:
-        num_processes: 4
+        num_processes: 8
       kafka-ucr-static-ccs_record_cases:
-        num_processes: 4
+        num_processes: 8
       kafka-ucr-static-ccs_record_cases_monthly:
-        num_processes: 4
+        num_processes: 8
       kafka-ucr-static-child_cases_monthly:
-        num_processes: 1
+        num_processes: 8
       kafka-ucr-static-child_health_cases:
-        num_processes: 4
+        num_processes: 8
       kafka-ucr-static-person_cases:
-        num_processes: 4
+        num_processes: 8
       KafkaDomainPillow:
         num_processes: 1
       LedgerToElasticsearchPillow:
         num_processes: 1
       SqlSMSPillow:
-        num_processes: 1
+        num_processes: 2
       UnknownUsersPillow:
         num_processes: 1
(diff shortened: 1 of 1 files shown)


fab/environments.yml
```diff
@@ -166,39 +166,39 @@ icds:
...
+        num_processes: 8
       kafka-ucr-static-ccs_record_cases:
-        num_processes: 4
+        num_processes: 8
       kafka-ucr-static-ccs_record_cases_monthly:
-        num_processes: 4
+        num_processes: 8
       kafka-ucr-static-child_cases_monthly:
-        num_processes: 1
+        num_processes: 8
```
snopoke, line 183: Not sure why this was only 1, presumably just an oversight.

fab/environments.yml
```diff
@@ -166,39 +166,39 @@ icds:
...
       kafka-ucr-static-person_cases:
-        num_processes: 4
+        num_processes: 8
       KafkaDomainPillow:
         num_processes: 1
       LedgerToElasticsearchPillow:
         num_processes: 1
       SqlSMSPillow:
-        num_processes: 1
+        num_processes: 2
```
snopoke: @gcapalbo bumped sms pillow as well

fab/environments.yml
```diff
@@ -165,40 +165,45 @@ icds:
         num_processes: 1
       CaseSearchToElasticsearchPillow:
```
emord: We should just remove case search pillow. I don't htink it will ever be used on icds

fab/environments.yml
```diff
@@ -165,40 +165,45 @@ icds:
...
+    '10.247.24.30': # pillow1
+      CaseToElasticsearchPillow:
         num_processes: 4
+      FormSubmissionMetadataTrackerPillow:
+        num_processes: 2
+      kafka-ucr-static-forms-A:
+        num_processes: 8
+      kafka-ucr-static-forms-B:
+        num_processes: 8
+      kafka-ucr-main:
```
emord: I'm not sure that we need this pillow in its current form.

We should probably have a new pillow that ignores any icds-cas submissions since all UCRs should be static for icds-cas. will confirm with @sheelio when he comes back from vacation

emord (commented):
There's a lot of bumping non-ucr queues. We'll need to deploy https://github.com/dimagi/commcare-hq/pull/16020 before we can bump them. I meant to deploy that after couch 2, but I've been on interrupt. I can make that a priority today if you want me to deploy.

Also this adds more than 30 processes. and makes pillow1 have 35 processes. are you concerned that this may be adding too much load at once? Right now pillow0 tops out at around ~13 load and pillow1 ~8 load. I'm particularly worried about pillow1 as it handles couch. We may need more resources and we should likely start deploying these in smaller increments

fab/environments.yml
```diff
@@ -165,40 +165,45 @@ icds:
...
       GroupPillow:
         num_processes: 1
       GroupToUserPillow:
         num_processes: 1
-      kafka-ucr-main:
-        num_processes: 1
       kafka-ucr-static-awc-location:
-        num_processes: 1
-      kafka-ucr-static-cases:
         num_processes: 4
```
emord: Awc-location only needs one process. it only listens to locations and it's totally caught up. having 4 is overkill

fab/environments.yml
snopoke: There is currently 1 non static UCR datasource

fab/environments.yml
snopoke: hmm, perhaps that graph is misleading

snopoke:
@emord I've changed it to only adjust UCR pillows and also re-evaluated the ones that need changing. Re the added load, I think this should still be OK. `pillow1` CPU and memory usage is generally quite low and with moving some off `pillow0` the net effect isn't huge.

fab/environments.yml
```diff
@@ -165,34 +165,37 @@ icds:
         num_processes: 1
       CaseSearchToElasticsearchPillow:
         num_processes: 1
-      CaseToElasticsearchPillow:
-        num_processes: 1
```
snopoke: removed this entirely

emord: Whoa. I think we should only remove `CaseSearchToElasticSearchPillow` If we remove `CaseToElasticsearchPillow` exports or reports such as case list won't work as they are dependant on that

snopoke: thanks, fixed: 972bcc5

PR #276, opened 2017-06-07
Increase concurrency on icds reminder case update worker

gcapalbo:
@emord 
buddies @sravfeyn @snopoke


emord (approved):

emord:
I think @snopoke has already started the deploy to NIC

gcapalbo:
ok, that's ok, i'm likely going to have to deploy in the next couple days at some point again

PR #277, opened 2017-06-08
Update environments.yml

benrudolph:
@gcapalbo the switch to the saved exports queue with only 3 workers is causing a fairly significant slowdown. bumping this up to 4


gcapalbo:
:+1:

PR #285, opened 2017-06-13
Update Celery Configuration

gcapalbo:
The first commit is a refactor which lets us consolidate all of our supervisor celery configuration files into one. Going forward, adding a worker consuming from a new queue will just require updating environments.yml.

The second commit allows us to add multiple workers on the same machine. This is most advantageous when using gevent pooling, and is likely something we will need to do for large scale.

Also, this makes all the workers use the bash runner. About half of our workers are using it now already and I think it's best if all use it.

Will be testing this on staging today, hold off on merge for now.

@sravfeyn @snopoke 
@benrudolph fyi


snopoke (approved):
Awesome

fab/operations/supervisor.py
```diff
@@ -57,34 +57,49 @@ def _get_celery_queues():
...
-        'saved_exports_queue':          ['supervisor_celery_saved_exports_queue.conf'],
-        'ucr_queue':                    ['supervisor_celery_ucr_queue.conf'],
-        'ucr_indicator_queue':          ['supervisor_celery_ucr_indicator_queue.conf'],
-        'email_queue':                  ['supervisor_celery_email_queue.conf'],
-        'repeat_record_queue':          ['supervisor_celery_repeat_record_queue.conf'],
-        'logistics_reminder_queue':     ['supervisor_celery_logistics_reminder_queue.conf'],
-        'logistics_background_queue':   ['supervisor_celery_logistics_background_queue.conf'],
-        'async_restore_queue':          ['supervisor_celery_async_restore_queue.conf'],
-        'flower':                       ['supervisor_celery_flower.conf'],
-    }
```
benrudolph, line 79: 😂

benrudolph (approved):
yess wanted this for so long!

fab/operations/supervisor.py
```diff
@@ -57,34 +57,49 @@ def _get_celery_queues():
...
+            'pooling': pooling,
+            'max_tasks_per_child': max_tasks_per_child,
+        })
+
+        for worker_num in range(num_workers):
+            params.update({
+                'worker_num': worker_num,
+            })
+
+            _rebuild_supervisor_conf_file(
```
emord: This creates a celery file file for every queue correct? Pillows currently create one large file with every process in it. It might be worth having them use the same format.

I don't really have an opinion on which is better to use. happy to add it to my backlog if we think multiple files are better

gcapalbo: yeah it creates one config file per worker as it was doing before. i don't really feel strongly on which is better - i don't think there's many pros/cons to either approach, besides the fact that it was easier to do it this way for celery because of how we start and stop the workers, so i think will be easier to maintain.

keeping pillows the way they are is probably also fine if it's easier to maintain that way for that process group.

gcapalbo:
now depends on https://github.com/dimagi/commcare-hq/pull/16610

gcapalbo:
tested on staging, am good with this now

fab/environments.yml
```diff
@@ -1,5 +1,24 @@
...
+#       max_tasks_per_child: 5  # Optional, default 50; only applicable for prefork pooling (corresponds to maxtasksperchild worker command line arg)
+#       num_workers: 2          # Optional, default 1; the number of workers to create consuming from this queue on this host
+#
+# The above celery configuration would create 2 celery workers with max concurrency of 4
+# using prefork pooling and a max tasks per child setting of 5, which consume from the
+# 'queue_name' queue and both would reside on the 'host_name' machine.
+#
+# It's ok to create workers on multiple machines as well. For example, adding 'queue_name'
+# under a 'host_name_2' section would also create workers on the 'host_name_2' machine
+# consuming from the same 'queue_name' queue.
```
snopoke, line 20: nice

snopoke (approved):

PR #295, opened 2017-06-20
Add workers to consume from prime_restore_queue

gcapalbo:
(If we decide to add a new queue for the task)
@calellowitz @snopoke @benrudolph


fab/environments.yml
```diff
@@ -79,6 +79,9 @@ softlayer:
       background_queue:
         concurrency: 4
         max_tasks_per_child: 1
+      prime_restore_queue:
```
gcapalbo, line 82: also if these tasks are mostly i/o bound we can set `pooling: gevent` here

snopoke:
I'd be in favor of killing this feature. It was only supposed to be a stop gap until async restore was released. But fine to do this for now until we can determine if we can actually remove it or not.

fab/environments.yml
snopoke: Definitely IO bound.

PR #296, opened 2017-06-20
add new storage devices to es2/es3

sravfeyn:
@emord

### DIFF
--- a/fab/inventory/icds
+++ b/fab/inventory/icds
@@ -71,7 +71,7 @@ hostname='es1'
 
 [es2:vars]
 elasticsearch_node_name=es2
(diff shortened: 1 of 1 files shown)


fab/inventory/icds
```diff
@@ -71,7 +71,7 @@ hostname='es1'
 
 [es2:vars]
 elasticsearch_node_name=es2
-devices='["/dev/sdb"]'
+devices='["/dev/sdb", "/dev/sde"]'
```
javierwilson: sdc and sdd are also used, only not added here, i think we should

sravfeyn: Updated

javierwilson (approved):

PR #298, opened 2017-06-20
specify softlayer pillows by machine

nickpell:
@emord 

The other environments specify pillow machines by IP address, but would alias work too?


emord:
@nickpell you should use IP addresses. I ran into this as well. background is here https://github.com/dimagi/commcare-hq-deploy/pull/246

PR #300, opened 2017-06-21
Update config on messaging celery workers for prod and softlayer

gcapalbo:
Switches them to use gevent pooling so we can increase processing power a bit.
@kaapstorm

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -66,14 +66,19 @@ softlayer:
         max_tasks_per_child: 1
     '10.162.36.194':  # celery1
       reminder_case_update_queue:
-        concurrency: 4
+        pooling: gevent
+        concurrency: 5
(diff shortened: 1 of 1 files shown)


fab/environments.yml
```diff
@@ -273,13 +278,18 @@ production:
...
-        concurrency: 8
+        pooling: gevent
+        concurrency: 5
+        num_workers: 2
       reminder_rule_queue:
         concurrency: 1
         max_tasks_per_child: 1
       sms_queue:
+        pooling: gevent
         concurrency: 8
```
gcapalbo, line 293: leaving this at 8 because we have one sms gateway that doesn't allow more than 8 concurrent connections, so i need to make a code update before we increase this on prod

kaapstorm (approved):

PR #302, opened 2017-06-21
beefy new celery server

emord:
@nickpell you can use this for the inventory to to the new celery server for enikshay. just needs to be updated with IP


nickpell:
nice, thanks for this

PR #306, opened 2017-06-23
Proxy port host for Formplayer

benrudolph:
@javierwilson either we've never error'd on formplayer on icds, or this is broken. guessing it's the latter. hold off on merge until i confirm this is working. these params are currently live on icds and web apps working. shouldn't be long until we get some errors


benrudolph:
cc: @orangejenny

orangejenny:
don't have good context on this, going to leave for @javierwilson

javierwilson (approved):

benrudolph:
hmm will have to look into this more. still didn't fix the issue

PR #309, opened 2017-06-26
Removed old pillows

emord:
will likely want to bump case-sql to 24 partitions this week-ish and add 8 process to kafka-ucr-static-cases. but as of now it catches up quickly

https://github.com/dimagi/commcarehq-ansible/pull/819

@calellowitz


calellowitz (approved):
not merging because of tag on linked pr but looks good

PR #311, opened 2017-06-30
moving scripts to check servers/services here

javierwilson:
i'm moving this here as fabric seems to be a better tool for the job.
still at an early stage, it works by running:
$ fab icds check_status

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -35,7 +35,7 @@ import pipes
 from distutils.util import strtobool
 
 from fabric import utils
-from fabric.api import run, roles, execute, task, sudo, env, parallel
+from fabric.api import run, roles, execute, task, sudo, env, parallel, local
 from fabric.colors import blue, red, magenta
 from fabric.context_managers import cd
 from fabric.contrib import files, console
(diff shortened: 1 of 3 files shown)


fab/fabfile.py
```diff
@@ -881,3 +891,16 @@ def reset_pillow(pillow):
     staticfiles.update_manifest,
     release.clean_releases,
 ]
+
+@task
+def check_status():
+    env.user = 'ansible'
+    env.sudo_user = 'root'
+    env.password = getpass('Enter the password for then ansbile user: ')
```
dannyroberts: then => the

dannyroberts (commented):
Hey @javierwilson, for my own context, where are these scripts being moved from?

fab/fabfile.py
javierwilson: good catch, i copy-and-pasted from L726 so fixing there too.

javierwilson:
I started writing a ping.yml for ansible, to write a kind of "check_services" at the OS-level not HQ dependant.  Since ansible is not really equipped to do the job, I'm doing this here instead.

dannyroberts (approved):

PR #312, opened 2017-07-06
new pillow machine for prod

javierwilson:


fab/environments.yml
```diff
@@ -311,7 +311,14 @@ production:
         concurrency: 4
         max_tasks_per_child: 1
   pillows:
-    '*':
+    '172.24.32.50':
```
emord, line 314: I think you can use `'hqpillowtop0'`/ `'hqpillowtop1`' for prod, but not sure. could you include a comment like # hqpillowtop1 or something slike we've done in the other envs>

PR #316, opened 2017-07-07
use hostname instead of ip address

emord:
@javierwilson @mkangia @dannyroberts 

Just hotfixed this on prod. Not sure why this didn't work, but pillows weren't running for the past 4.5 hours. I think IP addresses should always work, but it looks like they don't on prod, and hostnames don't work on softlyaer/icds, so it's pretty confusing.

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -323,14 +323,14 @@ production:
         concurrency: 4
         max_tasks_per_child: 1
   pillows:
-    '172.24.32.50':
+    hqpillowtop1:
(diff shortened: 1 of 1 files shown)


javierwilson:
odd. but in the inventory in softlayer and icds we use IP numbers  (not in prod), i am not sure it has anything to do with it.

emord:
@javierwilson yeah i was mistaken https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/utils.py#L38 only looks for host name

PR #322, opened 2017-07-12
increase pg connections on pg0 & pg3

snopoke:
Also decrease on es1

This will require restarting the DB's so should wait till off peak


snopoke:
@dimagi/scale-team

javierwilson (approved):

emord (approved):

emord:
I changed es1 back to 800
2017-07-12 19:02:55.278 GMT,,,5960,,5966725b.1748,4,,2017-07-12 19:02:51 GMT,,0,FATAL,22023,"hot standby is not possible because max_connections = 400 is a lower setting than on the master server (its value was 800)",,,,,,,,,""

emord:
these are deployed

PR #324, opened 2017-07-12
Split async restore worker to two workers on prod

gcapalbo:
https://app.datadoghq.com/notebook/9666/Prod-Celery-memory-usage

Looks like we have some capacity on hqcelery0 but hqcelery1 is continuously running out of memory. The async restore child worker processes each take up about 10% memory usage of the machine so this splits it into two workers on two different machines. Also reduced the max tasks per child setting to help reduce effects of memory leaking.

@dannyroberts

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -303,6 +303,9 @@ production:
         max_tasks_per_child: 5
       ucr_indicator_queue:
         concurrency: 1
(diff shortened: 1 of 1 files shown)


snopoke (approved):

emord (approved):

dannyroberts:
👍

PR #325, opened 2017-07-12
add 8G swap to all production celery machines

dannyroberts:
I chose 8G because it's equal to the total amount of RAM. Assuming we'll continue our tendency of using all machines to the max, this means that when we restart celery processes and the number of processes temporarily doubles, we'll still be well within the limits of RAM + swap so the OS won't be forced to use its oom killer. In addition, all three VMs have large drives of which they're currently using < 30%, so reserving 8G of disk for swap shouldn't be an issue.

### DIFF
--- a/fab/inventory/production
+++ b/fab/inventory/production
@@ -31,6 +31,7 @@ hqcelery2.internal-va.commcarehq.org
 
 [celery:vars]
 newrelic_app_name=celery
+swap_size=8G
 
 [pillowtop]
 hqpillowtop0.internal-va.commcarehq.org


emord:
You should make sure that swappiness is set to either 1 or 0 on the celery machines. I think default is 60

emord:
I think swap tends to be a parititon in linux right? Would this take downtime to provision/repoartion the drive? or can we use a file on the current partition

dannyroberts:
i think ansible takes care of swappiness: https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/swap/tasks/main.yml#L48-L54

As for it being on a linux partition, maybe i don't understand what I'm doing when I make swap files, but I would have said no. My understanding is that you literally just make an 8GB file and then designate it as swap, and linux starts using that file as swap.

emord (approved):
Nice. I have only ever done this as a separate partition. didn't realize you could use a file

PR #326, opened 2017-07-13
Add hqdb1 to prod inventory

kaapstorm:
@snopoke, does this look right? 

(I can confirm secondary storage is /dev/xvdb)

### DIFF
--- a/fab/inventory/production
+++ b/fab/inventory/production
@@ -11,6 +11,7 @@ newrelic_app_name=django
 
 [postgresql]
 hqdb0.internal-va.commcarehq.org hot_standby_server=hqstandby0.internal-va.commcarehq.org datavol_device=/dev/xvdb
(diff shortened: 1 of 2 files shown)


kaapstorm:
@javierwilson, Simon has a lot on his plate right now. Do you know whether I need to give more config for the new DB machine, or will the defaults be OK? -- I noticed the ICDS DB machines have a lot of vars set, and each of them has a hot standby.

snopoke (approved):
looks good. I don't think you need any of the other vars.

kaapstorm:
"set datadog_integration_cloudant=true for db0 on india" -- do you mean "on prod", @snopoke ?

javierwilson (approved):

javierwilson:
no standby for hqdb1?

snopoke:
@kaapstorm no I mean on India since I removed that var from the `postgresql` group it needs to be set on the host in the inventory

PR #329, opened 2017-07-14
no more load scaling celery

emord:
this just doesn't work...

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -192,41 +192,32 @@ icds:
       async_restore_queue:
         concurrency: 4
       ucr_indicator_queue:
-        concurrency: 16
-        celery_loader: 'corehq.util.celery_utils.LoadBasedLoader'
+        concurrency: 12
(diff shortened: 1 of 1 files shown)


gcapalbo:
too bad this didn't work out. thanks for all your work experimenting with this

snopoke:
🙁 I'd be interested to hear more about the specifics of why this didn't
work

emord:
@snopoke https://github.com/dimagi/commcare-hq/pull/17019 goes into a bit more detail of my last attempt to fix this.

Essentially celery (although more likely billiard (the pooling library)) didn't really listen when we say `scale_down`. it would leave workers in some idle state but not remove them from the pool. This meant that nothing was being processed so i reverted to suing a cron to bring workers up and down.

I'd be interested in attempting to fix this again because I think it could be a useful utility, but we don't have a lot of cycles right now and it seems like this feature is unsupported and undocumented by celery

PR #330, opened 2017-07-14
Mk enikshay add groups

mkangia:
Took hints using icds inventory

Highlights:
1. Added redis on es0 since that looked the one machine which could support it
2. added rabbitmq on kafka machine since that was running celery as well
3. not sure whats the difference between touchforms and formplayer is here, so added both


javierwilson (approved):

mkangia:
FYI
@snopoke

PR #335, opened 2017-07-19
Remove OPM pillow

orangejenny:
Merge with https://github.com/dimagi/commcare-hq/pull/17093

@esoergel


emord:
This doesn't actually need to wait on that PR. The only thing this changes is that it won't process any new changes using that pillow, which sounds like an ok thing to do since we don't care about it anymore

esoergel (approved):

PR #337, opened 2017-07-25
Changed ucr indicator queue to gevent.

emord:
This moves the ucr indicator queues to gevent. I kept the same # of workers because we'll still want to take advantage of all the cores, but it looks like there's more overhead with person cases. Even 16 workers are only taking ~25% cpu at most (which is where 4 came from)

@dimagi/scale-team


emord:
Alternatively could bump up the number of workers and keep it at prefork. We don't really have any memory issues on NIC

gcapalbo (approved):

emord:
This isn't working like I would have expected. I"m pretty confused by the behavior of this right now. will look again tomorrow

PR #338, opened 2017-07-25
Fixed migrations exists check.

emord:
This has been killing me recently. I think this should fix it, but I'll deploy staging again today with this to verify.

@nickpell @benrudolph


benrudolph (approved):
thanks i realized a while ago that it wasn't working, but haven't gotten a chance to fix.

emord:
This still runs migrate_multi, but at least it doesn't stop pillows or celery tasks which can take an annoying amount of time

PR #339, opened 2017-07-25
Run git gc in the background.

emord:
@dannyroberts @nickpell

### DIFF
--- a/fab/operations/release.py
+++ b/fab/operations/release.py
@@ -348,7 +348,7 @@ def mark_last_release_unsuccessful():
 
 def git_gc_current():
     with cd(env.code_current):
-        sudo('git gc &')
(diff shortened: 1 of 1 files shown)


emord:
hmm. this doesn't work for some reason...

benrudolph:
in other situations where we wanted to do this, we used the `at` command: https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/operations/db.py#L19

emord:
@benrudolph I think that that is still synchronous right? I was hoping to run `git gc` totally in the background so that it returns immediately and goes on to restarting services

emord:
Was using this incorrectly before. fixed

benrudolph (approved):

PR #340, opened 2017-07-25
Remove worker gossip

emord:
@gcapalbo I think this is what is causing those missed heartbeats. I don't think we use this unless this is how celery beat sends periodic tasks to workers http://docs.celeryproject.org/en/latest/whatsnew-3.1.html#gossip-worker-worker-communication


gcapalbo (approved):
thanks, yeah i don't think we use that at all right now

PR #342, opened 2017-07-28
More background queue processes

emord:
The background queue is pretty backed up on icds. Ideally we should probably re-evaluate what gets sent there, but adding this to celery1 seems fine

@gcapalbo @dimagi/scale-team @nickpell

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -199,6 +199,9 @@ icds:
     '10.247.24.31': # celery1
       ucr_indicator_queue:
         concurrency: 16
+      background_queue:
+        concurrency: 4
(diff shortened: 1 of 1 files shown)


benrudolph:
huh interesting i would not have expected the background queue to be hit hard on icds. what is the killer task?

emord:
@benrudolph I was looking into the ticket about the user export and noticed that there were a lot of rebuild export tasks in the background queue

gcapalbo (approved):

benrudolph:
interesting, would've thought they were on the `saved_export_queue`. thanks

emord:
yeah. i thought that was strange too

PR #344, opened 2017-08-02
no couchdb (only couchdb2)

javierwilson:
@emord  there's no need for couchdb now we use couchdb2, right?


snopoke (approved):

emord:
Yeah. once we migrate swiss to couch2 we can get rid of the couchdb role

PR #345, opened 2017-08-02
parallel supervisor

emord:
This will allow parallel supervisor services config generation. Before the workflow was generate all celery configs, generate all django configs,  ... This changes it to one step of generating all configs on every server at one time.

I think this would be a lot of faster if we moved it away from a manage.py command. Those commands are so slow to startup

@benrudolph @biyeun

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -802,7 +802,7 @@ def silent_services_restart(use_current_release=False):
 @task
 def set_supervisor_config():
     setup_release()
-    supervisor.set_supervisor_config()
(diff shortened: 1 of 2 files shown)


benrudolph (approved):
nice! this doesn't actually doing any of the restarting correct? it just runs the mgmt command to generate the conf. have wanted this for a while

emord:
Yep. It's really been killing me with these ICDS deploys. since we have so many workers on celery0 the celery generation takes forever. Should see some speedups everywhere else as well (anything that's not a monolith)

PR #346, opened 2017-08-02
add queue for reprocessing form submissions

snopoke:
buddy @orangejenny 

cf https://github.com/dimagi/commcare-hq/pull/17323


orangejenny (approved):

calellowitz:
Is there a reason this is only on ICDS

snopoke:
Just for now, once it's running smoothly we can deploy it to the other environments.

fab/operations/supervisor.py
```diff
@@ -188,6 +189,11 @@ def set_pillow_retry_queue_supervisorconf():
         _rebuild_supervisor_conf_file('make_supervisor_conf', 'supervisor_pillow_retry_queue.conf')
 
 
+def set_submissions_reprocessing_queue_supervisorconf():
+    if 'submission_reprocessing_queue' in _get_celery_queues():
+        _rebuild_supervisor_conf_file('make_supervisor_conf', 'supervisor_submission_reprocessing_queue.conf')
```
gcapalbo, line 194: @snopoke the deploy failed because supervisor_submission_reprocessing_queue.conf is missing from the repo. for now i'm going to comment the line out of `set_supervisor_config` to continue deploy.

also @emord pointed out that this should probably use `_check_in_roles` so it only gets deployed to the right machine

snopoke: sorry about that - fixed here: https://github.com/dimagi/commcare-hq-deploy/pull/350

PR #347, opened 2017-08-02
maybe this'll make sentry work on icds

benrudolph:
@wpride @dannyroberts maybe this'll make sentry work on icds?


dannyroberts:
Any idea why production and india would work without this if ICDS doesn't?

emord:
@dannyroberts ICDS uses a proxy to communicate outside the cluster and we specify the proxy with environment variables. prod/softlayer don't need that

dannyroberts:
👍 gotcha thanks for the context

dannyroberts (approved):

PR #348, opened 2017-08-03
finishing up the enikshay for fab deploy

javierwilson:


fab/environments.yml
```diff
@@ -37,6 +37,110 @@ base:
...
+        num_processes: 1
+      CaseSearchToElasticsearchPillow:
+        num_processes: 4
+      kafka-ucr-static:
+        num_processes: 4
+      ReportCaseToElasticsearchPillow:
+        num_processes: 1
+      SqlSMSPillow:
+        num_processes: 1
+      UnknownUsersPillow:
```
benrudolph, line 129: if this is just for nikshay, we probably won't need this

fab/fabfile.py
```diff
@@ -203,6 +203,15 @@ def l10k():
 
 
 @task
+@task
```
benrudolph: nit: 2 `@task`

fab/inventory/enikshay
```diff
@@ -12,13 +12,13 @@
 [formplayer0] # formplayer, riak
 172.25.2.7
 
-[formplayer1] # formplayer
+[formplayer1] # formplayer, riak
```
benrudolph, line 15: interesting that riak is going to be sharing resources. do you know how much RAM will be on that machine?

javierwilson: 12G, how much memory should we use for formplayer_memory ?

benrudolph: i think roughly splitting half for riak and half for formplayer like you did is likely fine. i just know nikshay heavily heavily uses formplayer. so far formplayer hasn't proven to be memory constrained

mkangia (approved):

fab/environments.yml
mkangia: @javierwilson i am not sure about removing this.
checking with @emord.

emord: We removed it for ICDS. should be ok to remove here.

It's only useful for forms that get submitted without a username (like some custom call via the api or something) which I don't think will happen with custom projects.

PR #354, opened 2017-08-16
add enikshay proxy to enikshay

emord:


emord:
@javierwilson @gcapalbo @mkangia 

Deploying proxy with this will probably change the ssl cert that it uses. I'm not sure how the letsencrypt certs were set up, but we should get one of those for enikshay.commcarehq.org and remove the old one once that's set up

javierwilson (approved):

PR #357, opened 2017-08-17
reduce number of pillows on softlayer

nickpell:
If pillows aren't using too many resources after this change, will move them back to one machine

@emord


emord (approved):
1 will probably be pretty tight whenever people have something like a case import to do. would be pro keeping it at 2

PR #359, opened 2017-08-22
Don't double define the indicator queue

emord:
@nickpell @gcapalbo


nickpell (approved):

emord:
I checked on enikshay and the 16 line is the one that won originally

gcapalbo:
yeah if we needed to we can actually support multiple workers on the same machine using the `num_workers` param: https://github.com/dimagi/commcare-hq-deploy/blob/02c17b4effdc018bfc4bceb1d7eb16e5e6347df6/fab/environments.yml#L70

PR #361, opened 2017-08-22
optimize celery saved exports queue

snopoke:
This adds the `-O` flag to `saved_exports_queue`

Still need to move normal exports to their own queue.


snopoke:
Testing this on a case history export for ICDS over 1K cases with 40K actions and 100 columns in each of the case and case history tables the timing is as follows:

* unoptimized: 124s
* optimized: 59s

This would still take 28 days to run over all person cases (40391107)

dannyroberts (approved):

PR #371, opened 2017-09-01
Reduce concurrency for repeater queue to avoid db connection errors

mkangia:
Repeaters are failing silently on enikshay due to concurrency. This is probably due to high concurrency in repeater queue. Going to reduce to fairly low number for now and monitor it.

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -52,7 +52,7 @@ enikshay:
         concurrency: 2
       repeat_record_queue:
         pooling: gevent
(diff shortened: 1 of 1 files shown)


gcapalbo (approved):
it would be good to monitor the repeat record queue to make sure it doesn't get behind if we're lowering it

mkangia:
> it would be good to monitor the repeat record queue to make sure it doesn't get behind if we're lowering it

👍 

>

PR #375, opened 2017-09-07
india has no standby

javierwilson:
i don't think we need this https://github.com/dimagi/commcare-hq-deploy/pull/126 anymore

### DIFF
--- a/fab/inventory/softlayer
+++ b/fab/inventory/softlayer
@@ -15,6 +15,7 @@
 
 [db0:vars]
 datadog_integration_cloudant=true
+encrypted_root=/opt/data
 
 [es2]
 10.162.36.221 elasticsearch_node_name=es2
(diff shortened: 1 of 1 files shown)


gcapalbo (approved):
does this mean we're changing how we do standbys, or are we not doing them at all?

curious if there are any docs on the new processes

javierwilson:
Not really, not yet at least, the procedure to create the standby has been changed but only to make it easier (use less disk space) that's on the standby wiki page.

But this is just regarding backup scripts. We shouldn't add a host to the standby group unless it is a standby, otherwise we can use "not is_pg_standby and not hot_standby_server" to run tasks on pg masters with no standbys.

What has changed is the backup solution we use, I have to document it yet, i will do so earlier next week, I hope.

PR #376, opened 2017-09-07
Fix default_branch usage

gcapalbo:
(deploying staging was always deploying master)

@snopoke 
@orangejenny

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -218,18 +218,18 @@ def _setup_env(env_name, force=False, default_branch=None):
 
 
 def _confirm_branch(default_branch=None):
-    if env.code_branch != 'master':
+    if env.code_branch == 'master':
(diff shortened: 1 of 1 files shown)


fab/fabfile.py
```diff
@@ -218,18 +218,18 @@ def _setup_env(env_name, force=False, default_branch=None):
 
 
 def _confirm_branch(default_branch=None):
-    if env.code_branch != 'master':
+    if env.code_branch == 'master':
         if default_branch:
             env.code_branch = default_branch
```
millerdev, line 223: Seems odd that it's impossible to deploy `master` to staging. Looks like it's been that way for a long time, even before 390d47e810f5fd2d81980bb60a7cef0893721229, so I'm not saying it needs to change.

millerdev (approved):

PR #378, opened 2017-09-13
move formplayer onto machines with free mem

dannyroberts:
formplayer0 => proxy0
formplayer1 => kafka0

My plan is to ansible deploy just formplayer to the new machines (without making proxy changes) and then to rsync the respective data directories. (Hopefully the proxy preserves hashing order like I think it does.) Then I'll deploy the proxy to smoothly switch over to the new machines, and then we can merge this

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -40,7 +40,7 @@ base:
 enikshay:
   environment: enikshay
   formplayer_memory: "7000m"
-  newrelic_javaagent: True
+  newrelic_javaagent: False
   celery_processes:
     '*':
       celery_periodic:
--- a/fab/inventory/enikshay
+++ b/fab/inventory/enikshay
@@ -126,8 +126,8 @@ datavol_device=/dev/mapper/consolidated-data
 es0
 
 [formplayer:children]
-formplayer0
-formplayer1
+proxy0
+kafka0
 
 [touchforms:children]
 web0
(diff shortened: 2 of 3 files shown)


gcapalbo:
do we have a sense for how running formplayer will affect the other services running on those machines during peak times?

if not, would it be better to just add an additional instance of formplayer on those machines instead of moving them, so that we have 4 total instead of just these 2?

dannyroberts:
I think we don't want to leave formplayer on the current machines because nearly all the memory is taken up by riak (it's not a matter of formplayer being over worked—it simply can't get the full amount of physical RAM it's asking for).

I'm not too concerned about proxy0 and kafka0; they're both massively oversized (e.g 50+GB RAM free each), at least in terms of memory, and formplayer seems to have a pretty low footprint resource-wise.

Does that help @gcapalbo, or do you think there's reason for concern?

gcapalbo (approved):
ok, sounds ok

mkangia (approved):
I am sure you have it in your steps, but just flagging that Its important to do a QA once that change is made to ensure things are working well.
@javierwilson added something to nginx for formplayer to work correctly, do we need to add anything for new machines then?

Good luck!

dannyroberts:
The switch-over is all complete and everything seems to be workings smoothly. I copied over all the sqlite dbs as well, so this shouldn't even result in more syncs or anything (though that probably would have been fine).

PR #380, opened 2017-09-14
Reduce number of pillows.

emord:
@proteusvacuum @nickpell @sravfeyn

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -110,19 +110,15 @@ enikshay:
       GroupToUserPillow:
         num_processes: 1
       kafka-ucr-main:
-        num_processes: 4
+        num_processes: 2
       KafkaDomainPillow:
         num_processes: 1
       LedgerToElasticsearchPillow:
         num_processes: 1
-      ReportXFormToElasticsearchPillow:
-        num_processes: 1
       CaseSearchToElasticsearchPillow:
         num_processes: 4
       kafka-ucr-static:
-        num_processes: 4
-      ReportCaseToElasticsearchPillow:
-        num_processes: 1
+        num_processes: 2
       SqlSMSPillow:
         num_processes: 1
       UnknownUsersPillow:


fab/environments.yml
```diff
@@ -110,19 +110,15 @@ enikshay:
       GroupToUserPillow:
         num_processes: 1
       kafka-ucr-main:
-        num_processes: 4
+        num_processes: 2
```
emord, line 113: This could probably be 1 actually. UCRs aren't processed here since you guys use the async queue. This just puts them in the queue

fab/environments.yml
```diff
@@ -110,19 +110,15 @@ enikshay:
...
       KafkaDomainPillow:
         num_processes: 1
       LedgerToElasticsearchPillow:
         num_processes: 1
-      ReportXFormToElasticsearchPillow:
```
emord, line 118: These aren't needed. I've been meaning to look into seeing if we can just trash them. It's some custom thing that's been hard to track down

emord:
I'm also wondering if you guys need the "Unknown users pillow". We turned it off for ICDS as we don't expect any submissions from non mobile workers

proteusvacuum:
I don't know enough about the unknown users pillow to say for sure. What does it populate and how would we use it?

emord:
@proteusvacuum unknown users pillow reads every form submitted and sees if the user is "known" if it is unknown it creates a new user in the users pillow. The workflow it uses is pretty broken and has been for years, but its on by backlog to fix it whenever i have the time (in like another 10 years)

I think it's ok to remove it if we don't expect forms to be coming from unknown users. We removed it from ICDS

proteusvacuum:
I'm not sure I understand how a user that isn't in our system would submit a form? However, we do have a lot of "system" forms that are submitted by the `system` user. I'm assuming that this user has already been added to the pillow, so sounds like it is safe to remove.

emord:
As long as you're using system then unknown user doesn't do anything. It ignores it from `WEIRD_USER_IDS` https://github.com/dimagi/commcare-hq/blob/master/corehq/pillows/user.py#L26

> I'm not sure I understand how a user that isn't in our system would submit a form? 

So this would be like if a user is submitting a form on behalf of another user. You can submit a form using one mobile worker, but change the metadata in the form/meta/username field to say that you're a different user

proteusvacuum:
~all our form submissions in the private sector app are submitted "on behalf of" using login-as. You can still only log in as users that exist in the system, so I assume that this was an issue because people were manually setting that field in past projects?

emord:
Added the unknonw user pillow back until we figure out if it's required

PR #381, opened 2017-09-15
Add more work memory to pg0

emord:
@dimagi/scale-team thought?

I think upgrading this will help the aggregation script go much faster. I think it will need to be paired with a PR to remove this from icds_public.yml in ansible, but just looking for feedback right now.

http://pgtune.leopard.in.ua/ says we should use 64 MB for web apps and 300 MB for a data warehouse (which this kind of is)

### DIFF
--- a/fab/inventory/icds
+++ b/fab/inventory/icds
@@ -14,7 +14,7 @@ hot_standby_server='10.247.24.16'
 postgresql_max_connections=800
 pgbouncer_max_connections=3500
 pgbouncer_default_pool=790
-pg_work_mem=32MB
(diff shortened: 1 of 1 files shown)


snopoke (approved):

emord:
to be merged with https://github.com/dimagi/commcarehq-ansible/pull/952 (already deployed)

snopoke (approved):

emord:
fyi looks like the aggregation was able to complete 1 to 1.5 hours earlier. may try a bump to 64 MB next week

PR #384, opened 2017-09-20
add formplayer0

snopoke:
buddy @millerdev


emord:
Any worries that we'll get some weird performance proxying equal requests to servers that have different specs?

snopoke:
Don't think so, also this is just temporary until hqformplayer0 is up and then I'll remove hqformplayer1

emord (approved):

PR #385, opened 2017-09-20
remove formplayer0

snopoke:


snopoke:
Still need to remove decommission the VM

PR #386, opened 2017-09-20
Reduce ucr indicator worker concurrency on celery0

gcapalbo:
Am reducing this to 4 from 12 and keeping the rest of the workers the same on this machine since this worker runs on a lot of other machines.

@emord

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -292,7 +292,7 @@ icds:
       async_restore_queue:
         concurrency: 4
       ucr_indicator_queue:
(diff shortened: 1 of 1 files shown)


emord (approved):
Would likely be fine to have it at 6, but this is a good start

gcapalbo:
ok, i'll keep an eye on it and increase if we can

emord:
Don't need to do it tonight. I"ll check it in the morning and will increase it if it makes it through the whole night with low usage

PR #399, opened 2017-10-26
double number of CaseSearchToElasticsearchPillow's

nickpell:
A first stab at addressing https://manage.dimagi.com/default.asp?262013

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -116,7 +116,7 @@ enikshay:
       LedgerToElasticsearchPillow:
         num_processes: 1
       CaseSearchToElasticsearchPillow:
-        num_processes: 4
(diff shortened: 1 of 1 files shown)


emord:
@calellowitz pillows are run on web0 which is probably going to conflict with your plans to run ucr indicator queues on there.

@nickpell do you think it's worth bumping the number of topics for case-sql to 16? (not that this should block you, just noting that this is the max number of process we'll be able to scale this pillow at the moment

nickpell:
@emord I'd say let's see if this is enough to bring down the average number of unprocessed cases, and if that works I'll stop there.

kaapstorm (approved):

PR #402, opened 2017-11-01
double enikshay UCR pillows

nickpell:
Running ```soft_delete_cases_with_no_person``` for prescriptions caused some UCR pillow backup:
https://app.datadoghq.com/dash/256236/change-feeds?live=false&page=0&is_auto=false&from_ts=1509477664295&to_ts=1509496962655&tile_size=m&tpl_var_env=enikshay&fullscreen=false

The next step is to run it for adherences, which will take several days to process.  Bumping the UCR pillow queues should help avoid backing up the queues during that time.  If doubling the number of pillows is not sufficient, then I can either redeploy with more pillows or pause the deletion process.

@sravfeyn @calellowitz @emord

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -113,7 +113,7 @@ enikshay:
       GroupToUserPillow:
         num_processes: 1
       kafka-ucr-main:
-        num_processes: 2
+        num_processes: 4
       KafkaDomainPillow:
         num_processes: 1
       LedgerToElasticsearchPillow:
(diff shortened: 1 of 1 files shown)


calellowitz:
Looks, good to me. Based on web0 load, i think this can handle the extra pillows without scaling back the new queues.

emord:
I'm amazed enikshay made it this far with only 2 processes. I think 8 should be fine. could also go ahead and bump up the queues to 16. as far as i know there's no reason to not do that

PR #406, opened 2017-11-06
ICDS deploy warning

snopoke:
As an added precaution

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -184,12 +184,14 @@ def softlayer():
 @task
 def icds():
     """www.icds-cas.gov.in"""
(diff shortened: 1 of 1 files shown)


fab/fabfile.py
```diff
@@ -184,6 +184,12 @@ def softlayer():
 @task
 def icds():
     """www.icds-cas.gov.in"""
+    # 05h30 - 20h30 IST
```
emord: did something similar for enikshay here https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/fabfile.py#L245

snopoke: thanks, updated

PR #409, opened 2017-11-07
Increase CaseToElasticsearchPillow for ICDS

emord:
@snopoke I think that the real answer here is to get case to elasticsearch pillow to use bulk apis. both get_cases to get cases in bulk from SQL and bulk update to ES


snopoke:
Cool, we already have all the code for that in the reindexer so I think it's just a case of refactoring to process cases in batches.

PR #390, opened 2017-10-05
postgres replication vars

snopoke:
Goes with https://github.com/dimagi/commcarehq-ansible/pull/962

buddy @nickpell

PR #411, opened 2017-11-08
PG replication vars for ICDS

snopoke:
Taken from https://github.com/dimagi/commcare-hq-deploy/pull/390

This has already been rolled out so good to commit to master.


emord:
what is spare?

snopoke:
good question. I guess it's just a `spare` replication slot (ie. not assigned). Probably not a good name for it though.

PR #412, opened 2017-11-08
Softlayer couch 234

dannyroberts:
Realized I never PR'd this last night.

### DIFF
--- a/fab/inventory/softlayer
+++ b/fab/inventory/softlayer
@@ -68,19 +68,19 @@ encrypted_root=/opt/data
 [couch1]
 10.162.36.218
 
-[couch2]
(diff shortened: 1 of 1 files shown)


fab/inventory/softlayer
```diff
@@ -139,6 +148,9 @@ riak15
 
 [couchdb2:children]
 couch1
+couch2
```
emord, line 151: Should the proxy get moved to one of these?

esoergel (approved):

fab/inventory/softlayer
dannyroberts: Yeah, good call. I'll do that next

PR #420, opened 2017-11-16
move read_inventory_file to fab/utils.py

dannyroberts:
so that it is easier to import from another package. (Importing fabfile.py causes print output which is only desirable when calling `fab` commands.)

### DIFF
--- a/fab/fabfile.py
+++ b/fab/fabfile.py
@@ -60,7 +60,6 @@ from .const import (
     PROJECT_ROOT,
 )
 from .exceptions import PreindexNotFinished
-from fab.utils import get_inventory
 from .operations import (
     db,
     staticfiles,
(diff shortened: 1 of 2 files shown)


fab/fabfile.py
```diff
@@ -70,12 +69,13 @@
...
-    clear_cached_deploy,
-    execute_with_timing,
     DeployMetadata,
     cache_deploy_state,
-    retrieve_cached_deploy_env,
+    clear_cached_deploy,
+    execute_with_timing,
+    read_inventory_file,
     retrieve_cached_deploy_checkpoint,
+    retrieve_cached_deploy_env,
```
kaapstorm, line 78: :joy: :heart: Changes like these let me know that I am not alone.

kaapstorm (approved):

PR #432, opened 2017-11-22
new ICDS inventory

snopoke:
@dimagi/scale-team 

- [ ] add riak IP addresses
- [x] configure postgres vars
- [ ] remote_couch_backup var - is this needed?


fab/environments.yml
```diff
@@ -378,6 +377,163 @@ icds:
       XFormToElasticsearchPillow:
         num_processes: 4
 
+icds-new:
+  environment: icds
```
sravfeyn: Should we name it different (e.g. `icds2`) so it doesn't conflict with `icds` if we deploy from local machine?

fab/inventory/icds-new
```diff
@@ -0,0 +1,602 @@
...
+10.247.164.47
+
+[formplayer0:vars]
+hostname='formplayer0'
+
+[patch0]
+10.247.63.132
+
+[proxy:children]
+web0
```
sravfeyn, line 483: Are we not having standalone proxy?

snopoke: We can but I don't think it's necessary

fab/environments.yml
snopoke: yea, this should be `icds-new`

PR #433, opened 2017-11-22
use old github library

nickpell:
there's still an issue to be worked out with the new one

### DIFF
--- a/fab/utils.py
+++ b/fab/utils.py
@@ -88,7 +88,7 @@ class DeployMetadata(object):
         repo.create_git_tag(
             tag=tag_name,
             message=msg,
-            object=self.deploy_ref,
+            obj=self.deploy_ref,
             type='commit',
             tagger=InputGitAuthor(
                 name=user.login,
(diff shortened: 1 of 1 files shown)


orangejenny:
Looks reasonable, though I don't have context...is this reverting a particular set of PRs? I thought it'd be a direct revert of https://github.com/dimagi/commcare-hq/pull/18232 but doesn't look like it.

fab/utils.py
```diff
@@ -11,7 +11,7 @@
 import re
 from getpass import getpass
 
-from github import Github, InputGitAuthor
+from github3 import login
```
orangejenny, line 14: Does github3 need to go back into requirements.txt, a la https://github.com/dimagi/commcare-hq/pull/18232/files#diff-444e80f30bd2f1e25b779966c9621b9bL90 ?

nickpell: yep: https://github.com/dimagi/commcare-hq/pull/18646

nickpell:
@orangejenny There were a few PRs that changed code, so I included the reverted commits from all of them

orangejenny (approved):

kaapstorm (approved):

PR #435, opened 2017-11-27
Ignore sdb volume which is 8GB

sravfeyn:
Refer to [fdisk output](https://docs.google.com/spreadsheets/d/1hN6Xo_tRNphDxMPTJE7OWydwOe_76ANlBtCapAKMi3s/edit#gid=428836263) to review

### DIFF
--- a/fab/inventory/icds-new
+++ b/fab/inventory/icds-new
@@ -474,6 +474,9 @@ hostname='rabbit0'
 10.247.164.47
 
 [formplayer0:vars]
+devices='["/dev/sdc"]'
+partitions='["/dev/sdc1"]'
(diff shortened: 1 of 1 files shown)


fab/inventory/icds-new
```diff
@@ -474,6 +474,9 @@ hostname='rabbit0'
 10.247.164.47
 
 [formplayer0:vars]
+devices='["/dev/sdc"]'
+partitions='["/dev/sdc1"]'
+datavol_device='/dev/mapper/consolidated-data'
```
sravfeyn, line 479: @snopoke I think you missed this, not intentional, right?

PR #440, opened 2017-11-27
Replace nodes for riak

mkangia:


millerdev:
Offlined with @mkangia. We're waiting for a response from @dannyroberts to a question about formplayer before making any changes to the riak cluster since it's best to make all changes at once rather than adding/removing one machine at a time. Hopefully we'll have that answer by tomorrow.

@mkangia feel free to re-open if this becomes relevant again.

PR #448, opened 2017-11-28
fixing periodic queue in icds-new

calellowitz:
@emord

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -390,9 +390,9 @@ icds-new:
       celery:
         concurrency: 8
         max_tasks_per_child: 5
-      celery_periodic:
-        concurrency: 4
-        server_whitelist: 10.247.24.19
+#      celery_periodic:
+#        concurrency: 4
+#        server_whitelist: 10.247.164.40
       pillow_retry_queue:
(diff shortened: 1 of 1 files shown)


fab/environments.yml
```diff
@@ -390,9 +390,9 @@ icds-new:
       celery:
         concurrency: 8
         max_tasks_per_child: 5
-      celery_periodic:
-        concurrency: 4
-        server_whitelist: 10.247.24.19
+#      celery_periodic:
+#        concurrency: 4
+#        server_whitelist: 10.247.164.40
```
gcapalbo, line 395: is this to prevent periodic tasks from running? note that if celery beat is still running the tasks will still be getting queued up.

i think you might just want to make sure that all the enqueueing processes are always stopped until go live (including celery beat, run_sms_queue, run_reminder_queue, queue_schedule_instances, run_submission_reprocessing_queue, run_pillow_retry_queue)

gcapalbo: (taking the conversation to https://github.com/dimagi/commcare-hq-deploy/pull/449#issuecomment-347560246)

PR #449, opened 2017-11-28
Remove other clery queues

emord:
@gcapalbo thanks for catching https://github.com/dimagi/commcare-hq-deploy/pull/448

Your comment didn't totally match up with the queues we have defined. is there an easy way to stop celery beat from running?

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -397,8 +397,8 @@ icds-new:
         concurrency: 1
       submission_reprocessing_queue:
         concurrency: 1
-      email_queue:
-        concurrency: 2
+      # email_queue:
+      #   concurrency: 2
     '10.247.164.41':
       ucr_queue:
         concurrency: 4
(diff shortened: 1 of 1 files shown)


gcapalbo:
@emord actually looking at the [fab code](https://github.com/dimagi/commcare-hq-deploy/blob/8b8d9504f7bcd393c4e6bc13e55fc0672c54096f/fab/operations/supervisor.py#L181-L199), it looks like the enqueueing processes i mentioned should also not be deployed if the worker is not deployed (though it would be good to doublecheck this works as intended). and the same should be true for [celery beat](https://github.com/dimagi/commcare-hq-deploy/blob/8b8d9504f7bcd393c4e6bc13e55fc0672c54096f/fab/operations/supervisor.py#L113-L118) which shouldn't be deployed if the periodic worker isn't deployed.

so i think the only workers that should need to be removed are:

- celery_periodic
- reminder_queue
- sms_queue
- submission_reprocessing_queue
- pillow_retry_queue

and the rest can be deployed and will just be idle unless you do something in the site manually, like initiate a web worker invitation which spawns an email task

emord:
Thanks for the explanation. Updated.

(for reference this is just temporary while we do performance testing)

PR #461, opened 2017-12-04
Remove r.js check before running build_requirejs

orangejenny:
@millerdev 

I suspect this is why the requirejs build step didn't run on the first deploy after requirejs was merged. That build step is running now; the various prod servers now have, and are serving, bundle.js files.

I'm not sure exactly why this would be a problem, since `collectstatic` should have made the `staticfiles/r.js` directory before this check runs, so it should have returned true even on that first deploy. But the bundle.js files weren't created on the first deploy, yet they have been created since then, without any other code changes or intervention, so I'm inclined to just remove this line, which was always meant to be temporary, and not dig any further.

### DIFF
--- a/fab/operations/staticfiles.py
+++ b/fab/operations/staticfiles.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import
 from fabric.api import roles, parallel, sudo, env
 from fabric.context_managers import cd
-from fabric.contrib import files
 
 from fab.utils import bower_command
 
@@ -77,8 +76,7 @@ def collectstatic(use_current_release=False):
         sudo('{}/bin/python manage.py collectstatic --noinput -v 0'.format(venv))
         sudo('{}/bin/python manage.py fix_less_imports_collectstatic'.format(venv))
         sudo('{}/bin/python manage.py compilejsi18n'.format(venv))
-        if files.exists('{}/staticfiles/r.js'.format(env.code_current)):
-            sudo('rm -f tmp.sh resource_versions.py; {}/bin/python manage.py build_requirejs'.format(venv))
+        sudo('rm -f tmp.sh resource_versions.py; {}/bin/python manage.py build_requirejs'.format(venv))
 
 
 @parallel


fab/operations/staticfiles.py
```diff
@@ -77,8 +76,7 @@ def collectstatic(use_current_release=False):
         sudo('{}/bin/python manage.py collectstatic --noinput -v 0'.format(venv))
         sudo('{}/bin/python manage.py fix_less_imports_collectstatic'.format(venv))
         sudo('{}/bin/python manage.py compilejsi18n'.format(venv))
-        if files.exists('{}/staticfiles/r.js'.format(env.code_current)):
-            sudo('rm -f tmp.sh resource_versions.py; {}/bin/python manage.py build_requirejs'.format(venv))
+        sudo('rm -f tmp.sh resource_versions.py; {}/bin/python manage.py build_requirejs'.format(venv))
```
millerdev, line 79: I don't really understand what this line is doing, so completely trusting you here. I'm interested in knowing why `resource_versions.py` is being removed (to be honest I don't even really know exactly what is in that file). Do you know?

millerdev (approved):

fab/operations/staticfiles.py
orangejenny: It's how we invalidate the CDN's caching. resource_versions.py is a massive dictionary generated during deploy (specifically the `resource_static` command) that maps every static file in HQ to a sha1 hash of its contents. The `static` template tag then looks up that hash when it generates a CDN link and uses it as the version, creating links like https://dnwn0mt1jqwp0.cloudfront.net/static/notifications/js/notifications_service.js?version=4db9b01  When code changes, the hash changes, so the CDN doesn't recognize the version and gets the new file from HQ.

Because the `build_requirejs` command writes some js files (one for each bundle, plus updating requirejs config files with bundle info), it also needs to set their hashes in resource_versions.py  I don't know exactly what the `rm` is for; I took this line from `_version_static` further up in this file because it also writes out resource_versions.py.

millerdev: Thanks for the explanation.

PR #462, opened 2017-12-05
Split pillows across servers

emord:
@snopoke I think eventually we'll want to totally change the way this is configured and maybe deployed. Originally I wanted to keep this close to celery config, but that may not be the best way to configure these.

Decided against figuring these out progamatically because a) These are stored in dicts so we don't have a guarantee that they keep order when iterating (though we could sort it on IP or something), and b) the supervisor config creation is run on each server and they only have access to their own config https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/utils.py#L45-L52

buddy @proteusvacuum

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -501,22 +501,36 @@ icds-new:
         num_processes: 1
     '10.247.164.39': # pillow2
       kafka-ucr-static-cases:
+        start_process: 0
         num_processes: 6
+        total_processes: 24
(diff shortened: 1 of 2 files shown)


snopoke (approved):

snopoke:
Agree that this is a bit clunky but it works for now. It would be good to validate the config though since it's quite easy to mess it up.

emord:
verified that this works on pillow3 on icds-new

PR #472, opened 2017-12-14
update pgbouncer pool size

snopoke:


sravfeyn:
I should have been more clear in my last PR, but are there any other parameters we need to update?

PR #476, opened 2017-12-15
increase max psql connections for sharded dbs

emord:
@dimagi/scale-team Made these changes for the performance testing

@calellowitz @snopoke I know you guys were doing some things iwth the standby set up. When i tried to deploy this I got an error start pgshard1standby because it didn't have any dbs. Is that expected?

```
root@pgshard1standby:/opt/data/postgresql/9.4/main# ls -lhtr
total 4.0K
drwxr-xr-x 2 postgres postgres 4.0K Dec 15 18:43 pg_log
```


emord:
Closing this since past 800 we were seeing consistent swapping on the shard dbs

PR #482, opened 2017-12-18
use correct drive names

snopoke:
I noticed that some of these don't match the output of `lvmdiskscan`.


sravfeyn (approved):

sravfeyn:
Thanks. This is probably because we cleaned up the extra 7GB disk that was there on these

PR #485, opened 2017-12-20
only use email command if email is working

czue:
don't merge until I confirm this works

alternatives (assuming this is the root problem):

- we could fail silently if email's not working in `preindex_everything` 
- we could fail more loudly in the fabfile if `preindex_everything` dies


snopoke (approved):

czue:
we'll have to remember to come back and update this once email is setup

snopoke:
I've added it to the tracker

PR #481, opened 2017-12-18
reenabling periodic queues

calellowitz:
@snopoke @emord @javierwilson @sravfeyn @dannyroberts 
This enables periodic queues on the new icds cluster. It will require a fab deploy before it takes effect

PR #486, opened 2017-12-20
re-enable periodic queue on icds

czue:
grepped for `CELERY_PERIODIC_QUEUE` and everything looks fine. we will run case update rules (seems fine) and attempt to send report emails (which will error, but I think that's ok?)

this will conflict with https://github.com/dimagi/commcare-hq-deploy/pull/481/ - I can update that PR once merged

### DIFF
--- a/fab/environments.yml
+++ b/fab/environments.yml
@@ -391,10 +391,10 @@ icds-new:
       celery_periodic:
          concurrency: 4
          server_whitelist: 10.247.164.40
-      # pillow_retry_queue:
-      #   concurrency: 1
-      # submission_reprocessing_queue:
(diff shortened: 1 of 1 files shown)


snopoke (approved):

gcapalbo:
if the sms gateway outage lasts past this coming monday, this task should also be commented: https://github.com/dimagi/commcare-hq/blob/master/custom/icds/tasks.py#L107-L111

czue:
thanks @gcapalbo . just checking - we could do that _on_ monday or would have to do it before monday?

gcapalbo:
@czue it would just need to be done before 9am IST on Monday

PR #487, opened 2017-12-21
squid at old es0

sravfeyn:


fab/inventory/icds-new
```diff
@@ -480,7 +480,7 @@ datavol_device='/dev/mapper/consolidated-data'
 hostname='formplayer0'
 
 [patch0]
-10.247.63.132
+10.247.24.16
```
dannyroberts, line 483: should there be a

```
[squid:children]
patch0
```

here or something? Just naively looking at the code (e.g. `deploy_http_proxy.yml`) it looks like that's the intention.

PR #488, opened 2017-12-22
Enable messaging on icds-new

gcapalbo:
The SMS tests I ran from web0 went through so it seems that machine is whitelisted already. While we're waiting for celery3 to be whitelisted, we can just run the sms queue worker from web0 so that messaging gets up and running.

@dimagi/scale-team


snopoke (approved):

gcapalbo:
replaced here: https://github.com/dimagi/commcarehq-ansible/pull/1193

